PR #266, opened 2021-12-01
fix: use correct override index for built-in iconset components

dpilch:
**Wait until after initial release to merge change.**

Built-in iconset components require the component type to be mapped to pascal calse. A new object was created in amplify-renderer to map the component type. This resulted in the [object equality comparison in studio-node](https://github.com/aws-amplify/amplify-codegen-ui/blob/4c82035c41b2ff015126e0f2751e616b5b7761fd/packages/codegen-ui/lib/frontend-manager-node.ts#L46) to fail because the objects were different references.


alharris-at:
You should be good to merge this now.

PR #279, opened 2021-12-06
ci: automate npm release

dpilch:
It is a little difficult to fully test this without publishing to NPM. We could make a fake release to fully test it out if we want to. I did confirm that `lerna publish from-git -y` will not publish if the HEAD commit does not have a tag so there is limited risk of accidentaly publishing.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,3 +33,22 @@ jobs:
           body_path: release_description.md
           draft: false
           prerelease: false
+
+  npm-release:
+    needs: github-release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Setup Node.js LTS
+        uses: actions/setup-node@v2
+        with:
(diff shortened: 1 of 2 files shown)


alharris-at:
How is this intended to interact w/ our release process? This would need to run after that workflow, right?

alharris-at:
Probably want to update docs as a part of this as well.

dpilch:
Oh yeah, I'll update the docs.

This would be the normal flow of create the release PR and then merging to main. After the GitHub release is created the workflow would then automatically publish to NPM.

.github/workflows/release.yml
```diff
@@ -33,3 +33,22 @@ jobs:
           body_path: release_description.md
           draft: false
           prerelease: false
+
+  npm-release:
+    needs: github-release
```
alharris-at, line 38: Ahh, this line is what I'd overlooked. Didn't realize you'd already built in the workflow serialization. Nice!

alharris-at (approved):

PR #286, opened 2021-12-07
chore: re-enable CLI util unit tests

alharris-at:


dpilch:
There aren't any unit tests in `amplify-cli/packages/amplify-util-adminui` so I'm not sure it will cover anything for us.

alharris-at:
> There aren't any unit tests in `amplify-cli/packages/amplify-util-adminui` so I'm not sure it will cover anything for us.

There seem to be a few, but it's pretty minimal. I'm mostly using this test now to ensure that I'm able to inject our library into CLI for testing. Next up will be setting up tests that rely on amplify dev to pull an app down and render.

dpilch:
Oh you are right, I missed them nested under `commands/utils/`.

dpilch (approved):

PR #292, opened 2021-12-08
ci: add npm pre-release

dpilch:
This will help CLI to easily test new changes to codegen without the need for `npm link`.

This change will publish a pre-release to npm for every change pushed to main that is:
* not a production release. (release commit)
* has functional changes to packages recognized by lerna
  * Ex. updating `.circleci/config.yml` will not make a pre-release, but changes `packages/codegen-ui/lib/studio-node.ts` will

The `--canary` option will look back one commit to determine if there are any changes that need to be published as a pre-release. If there are changes it will increase the patch version and append the commit hash to the version number before publishing. These changes are **NOT** kept in git. 

I have not found an easy way to look back further than one commit. This means if we rebase and merge to main and the HEAD commit does not make any changes recognized by lerna the pre-release would not be published. However, in most circumstances we use squash and merge so it is not a concern most of the time.

`--pre-dist-tag next` will publish the pre-release under the `next` tag keeping `latest` (default install) always pointing at production releases.

Example:
```
> lerna publish --pre-dist-tag next --canary -y

info cli using local version of lerna
lerna notice cli v4.0.0
lerna info canary enabled
lerna info Looking for changed packages since 2baaca9^..2baaca9

Found 2 packages to publish:
 - @aws-amplify/codegen-ui-react => 1.1.1-alpha.270+2baaca9
 - @aws-amplify/codegen-ui => 1.1.1-alpha.270+2baaca9
```

See lerna docs: https://github.com/lerna/lerna/tree/main/commands/publish

Note: their documentation is actually incorrect on the functionality of `--canary`. They state it makes a minor bump but that is in contradiction to their examples and my testing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -76,9 +76,12 @@ jobs:
       - run:
           name: NPM Publish
           command: >-
+            # lerna publish docs: https://github.com/lerna/lerna/tree/main/commands/publish
             if [[ `git log --format=%s -n 1` =~ ^chore\(release\):\ v.* ]];
+            # production release
             then npx lerna publish from-git -y;
-            else echo "No publish";
+            # pre-release
+            else npx lerna publish --pre-dist-tag next --canary -y;
             fi
 
 workflows:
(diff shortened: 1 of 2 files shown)


alharris-at (approved):
N.p. could you just add a bash comment in here to each line describing what they do?

dpilch:
I'll add a pre-release section to docs and some comments.

PR #297, opened 2021-12-09
fix: map prop to children prop for variant

yeung-wah:
*Issue #, if available:*

 map prop to children prop for variant

*Description of changes:*
*Revision3*
Moved getComponentTypeFromOverrideKey function to under StudioNode class

*Revision2*
In revision 1, missed to handle a case where the overrideKey in variant can be a nested children component(Flex.Flex[0].Flex[0].Button[0]), added a new function to parse overrideKey and get the corresponding component type

Added complex test using real world payload, the code for the nested children components which have the mapping should flip label prop to children prop, and display different text

Added test to not flip prop to children prop if the component is not defined in the mapping table

*Revision1*
Currently we are not remapping prop to children prop for variant. This change would remap predefined prop to children prop for variant. It is very similar to what we are doing with property binding. It will go through the variants section, and remap all the predefined prop to children for the specific primitive hierarchy based on this mapping https://github.com/aws-amplify/amplify-codegen-ui/blob/868867b0a3301e833849129855911c071caa9ba4/packages/codegen-ui-react/lib/primitive.ts#L66

Example:
```
{
  "id": "1234-5678-9010",
  "componentType": "Button",
  "name": "ComponentWithVariant",
  "properties": {
    "children": {
      "value": "ComponentWithVariant"
    }
  },
  "variants": [
    {
      "variantValues": {
        "variant": "success"
      },
      "overrides": {
        "Button": {
          "label": "ComponentWithVariantWithMappedChildrenProp",
          "fontSize": "15px"
        }
      }
    }
  ]
}
```

In memory it will turn into 
```
{
  "id": "1234-5678-9010",
  "componentType": "Button",
  "name": "ComponentWithVariant",
  "properties": {
    "children": {
      "value": "ComponentWithVariant"
    }
  },
  "variants": [
    {
      "variantValues": {
        "variant": "success"
      },
      "overrides": {
        "Button": {
          "children": "ComponentWithVariantWithMappedChildrenProp",
          "fontSize": "15px"
        }
      }
    }
  ]
}
```

and output
```
/* eslint-disable */
import React from "react";
import {
  EscapeHatchProps,
  Variant,
  getOverrideProps,
  getOverridesFromVariants,
} from "@aws-amplify/ui-react/internal";
import { Button, ButtonProps } from "@aws-amplify/ui-react";

export type ComponentWithVariantProps = React.PropsWithChildren<
  Partial<ButtonProps> & {
    variant?: "success";
  } & {
    overrides?: EscapeHatchProps | undefined | null;
  }
>;
export default function ComponentWithVariant(
  props: ComponentWithVariantProps
): React.ReactElement {
  const { overrides: overridesProp, ...rest } = props;
  const variants: Variant[] = [
    {
      variantValues: { variant: "success" },
      overrides: {
        Button: {
          fontSize: "15px",
          children: "ComponentWithVariantWithMappedChildrenProp",
        },
      },
    },
  ];
  const mergeVariantsAndOverrides = (
    variants: EscapeHatchProps,
    overrides: EscapeHatchProps
  ): EscapeHatchProps => {
    const overrideKeys = new Set(Object.keys(overrides));
    const sharedKeys = Object.keys(variants).filter((variantKey) =>
      overrideKeys.has(variantKey)
    );
    const merged = Object.fromEntries(
      sharedKeys.map((sharedKey) => [
        sharedKey,
        { ...variants[sharedKey], ...overrides[sharedKey] },
      ])
    );
    return {
      ...variants,
      ...overrides,
      ...merged,
    };
  };
  const overrides = mergeVariantsAndOverrides(
    getOverridesFromVariants(variants, props),
    overridesProp || {}
  );
  return (
    /* @ts-ignore: TS2322 */
    <Button
      children="ComponentWithVariant"
      {...rest}
      {...getOverrideProps(overrides, "Button")}
    ></Button>
  );
}

```

Since we are already taking care of children props in variant, it will follow the existing process, and display different text. 


packages/codegen-ui-react/lib/react-studio-template-renderer.ts
```diff
@@ -1296,5 +1302,31 @@ export abstract class ReactStudioTemplateRenderer extends StudioTemplateRenderer
...
+    }
+
+    this.component.variants.forEach((variant) => {
+      // loop through the keys in the dict. Ex. button, button.text
+      Object.keys(variant.overrides).forEach((overridesKey) => {
+        const childrenPropMapping = PrimitiveChildrenPropMapping[Primitive[overridesKey as Primitive]];
+        if (childrenPropMapping !== undefined) {
+          const propsInOverrides = variant.overrides[overridesKey];
+          // only remap if children prop is not defined in this particular overrides section
+          if (propsInOverrides.children === undefined && propsInOverrides[childrenPropMapping] !== undefined) {
```
alharris-at, line 1322: Even if propsInOverrides[childrenPropMapping] === undefined we probably want to map for consistency sake.

packages/codegen-ui-react/lib/react-studio-template-renderer.ts
```diff
@@ -1296,5 +1302,31 @@ export abstract class ReactStudioTemplateRenderer extends StudioTemplateRenderer
...
+   *
+   */
+  private mapSyntheticPropsForVariants() {
+    if (!isStudioComponentWithVariants(this.component)) {
+      return;
+    }
+
+    this.component.variants.forEach((variant) => {
+      // loop through the keys in the dict. Ex. button, button.text
+      Object.keys(variant.overrides).forEach((overridesKey) => {
```
alharris-at, line 1317: Instead of `Object.key` you can use `Object.entries` since you are accessing the values as well as the keys down below.

alharris-at (commented):
A few minor comments, but overall looks good to me. Could you also add a unit test to verify label _doesn't_ get remapped if it's not in your table?

packages/codegen-ui-react/lib/react-studio-template-renderer.ts
dpilch: This would prevent the option of using `children` in overrides on components that have children prop mapping because the `undefined` children prop mapping would overwrite `children`.

```js
"overrides": {
  "Button": {
    "children": "componentWithVariantsWithChildrenProp"
  }
}
```
Converts to:
```js
"overrides": {
  "Button": {
    "children": undefined
  }
}
```

yeung-wah: hmmm, childrenPropMapping here is referring to the predefined prop that we want to map to children prop(Ex label). For example, we need to remap label prop to children prop for button component
```
 "overrides": {
        "Button": {
          "label": "ComponentWithVariantWithMappedChildrenProp",
          "fontSize": "15px"
        }
      }
```
convert to 

```
 "overrides": {
        "Button": {
          "children": "ComponentWithVariantWithMappedChildrenProp",
          "fontSize": "15px"
        }
      }
```

if children prop is not defined(propsInOverrides.children) and label prop is defined(propsInOverrides[childrenPropMapping], then we will remap label prop to children prop. So if propsInOverrides[childrenPropMapping] is undefined, then we don need to do anything as the label prop is not there.

yeung-wah: > Even if propsInOverrides[childrenPropMapping] === undefined we probably want to map for consistency sake.
hmmm, childrenPropMapping here is referring to the predefined prop that we want to map to children prop(Ex label). For example, we need to remap label prop to children prop for button component

 "overrides": {
        "Button": {
          "label": "ComponentWithVariantWithMappedChildrenProp",
          "fontSize": "15px"
        }
      }
convert to

 "overrides": {
        "Button": {
          "children": "ComponentWithVariantWithMappedChildrenProp",
          "fontSize": "15px"
        }
      }
if children prop is not defined(propsInOverrides.children) and label prop is defined(propsInOverrides[childrenPropMapping], then we will remap label prop to children prop. So if propsInOverrides[childrenPropMapping] is undefined, then we don need to do anything as the label prop is not there.

packages/codegen-ui-react/lib/react-studio-template-renderer.ts
```diff
@@ -1296,5 +1302,52 @@ export abstract class ReactStudioTemplateRenderer extends StudioTemplateRenderer
...
+    });
+  }
+
+  /* Parse the override key to get the component type
+   *
+   * Example1: Flex.Flex[0].Flex[0].Button[0]
+   * Example2: Button
+   *
+   */
+  private getComponentTypeFromOverrideKey(overrideKey: string): string {
```
alharris-at, line 1340: Could you move this method to the [StudioNode](https://github.com/aws-amplify/amplify-codegen-ui/blob/main/packages/codegen-ui/lib/studio-node.ts) class as a `static` method, and ensure we've got unit tests covering it?

alharris-at (approved):

PR #302, opened 2021-12-10
feat: add min required library versions for generated components

alharris-at:
*Issue #, if available:*
https://app.asana.com/0/1201447141707771/1201500361005705/f

*Description of changes:*
Expose a utility class which will provide the CLI metadata on min required dependencies for generated components.


dpilch (approved):
An uncommon case, but should we use a semantic version range so that `@aws-amplify/ui-react@3.0.0` would be invalid?

dpilch (approved):

PR #303, opened 2021-12-10
chore: cherry-pick https://github.com/aws-amplify/amplify-codegen-ui/pull/302 into main

alharris-at:
Cherry-picking so that we can release with the supported version flag on Monday.


dpilch (approved):
When squash and merging make sure you use `feat: add min required ...` so that it will show in our changelog.

PR #307, opened 2021-12-14
feat: make prettier an optional dependency

dpilch:
Move prettier to optional dependency: https://docs.npmjs.com/cli/v8/configuring-npm/package-json#optionaldependencies

If prettier is not installed, do not run `prettier.format` on generated code.

We could also expose an option (something like `formatGeneratedCode`) and throw an error if `formatGeneratedCode` is `true` and prettier is not installed.

### DIFF
--- a/packages/codegen-ui-react/lib/react-studio-template-renderer-helper.ts
+++ b/packages/codegen-ui-react/lib/react-studio-template-renderer-helper.ts
@@ -35,8 +35,6 @@ import ts, {
   createProgram,
 } from 'typescript';
 import { createDefaultMapFromNodeModules, createSystem, createVirtualCompilerHost } from '@typescript/vfs';
-import prettier from 'prettier';
-import parserTypescript from 'prettier/parser-typescript';
 import path from 'path';
 import { ReactRenderConfig, ScriptKind, ScriptTarget, ModuleKind } from './react-render-config';
 
@@ -73,7 +71,7 @@ export function transpile(
       },
     }).outputText;
 
-    const componentText = prettier.format(transpiledCode, { parser: 'typescript', plugins: [parserTypescript] });
+    const componentText = formatCode(transpiledCode);
 
     /*
      * createProgram is less performant than traspileModule and should only be used when necessary.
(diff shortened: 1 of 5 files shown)


dpilch:
Integration test failures appear to be unrelated to this change. I am getting the same error on main.

packages/codegen-ui-react/lib/react-studio-template-renderer-helper.ts
```diff
@@ -186,3 +184,22 @@ export function bindingPropertyUsesHook(
 ): boolean {
   return isDataPropertyBinding(binding) && 'predicate' in binding.bindingProperties;
 }
+
+// optional import prettier
+export function formatCode(code: string): string {
+  let prettier = null;
+  let parserTypescript = null;
```
alharris-at, line 191: You can get rid of the `let's` here if you wrap the return inside your `try` block below, and may clean up the code a bit.

packages/codegen-ui-react/package.json
```diff
@@ -29,13 +29,15 @@
...
     "@aws-amplify/codegen-ui": "1.2.0",
     "@typescript/vfs": "~1.3.5",
-    "prettier": "2.3.2",
     "typescript": "~4.4.4"
   },
   "peerDependencies": {
     "react": "^17.0.2",
     "react-dom": "^17.0.2"
   },
+  "optionalDepdencies": {
```
alharris-at, line 38: Depdencies

packages/codegen-ui-react/lib/react-studio-template-renderer-helper.ts
dpilch: Good idea.

packages/codegen-ui-react/package.json
dpilch: Will fix

alharris-at (approved):

dpilch:
Moving to optional dependencies seems to have broken the combined package lock for cache key generation. I'll look into what is causing the issue.

dpilch:
Seems to actually have been a one-time issue with NPM.

PR #310, opened 2021-12-15
Operator type

dhandsar-aws:
*Issue #, if available:*

*Description of changes:*
Adding strong type for operator to accept only certain operator string values

### DIFF
--- a/packages/codegen-ui/lib/types/studio-types.ts
+++ b/packages/codegen-ui/lib/types/studio-types.ts
@@ -321,13 +321,18 @@ export type ConcatenatedStudioComponentProperty = {
 };
 
 /**
+ * These are the string values accepted by operator
+ */
+export type RelationalOperator = 'eq' | 'ne' | 'le' | 'lt' | 'ge' | 'gt';
+
+/**
  * Component property that represents a conditional expression
  */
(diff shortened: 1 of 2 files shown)


packages/codegen-ui/lib/types/studio-types.ts
```diff
@@ -320,14 +320,19 @@ export type ConcatenatedStudioComponentProperty = {
   concat: StudioComponentProperty[];
 };
 
+/**
+ * These are the string values accepted by operator
+ */
+export type Operator = 'eq' | 'ne' | 'le' | 'lt' | 'ge' | 'gt';
```
alharris-at, line 326: Could you update to a slightly more specific name, like 'ConditionalOperator'

dhandsar-aws: Good point, but to be specific - How about this name `Comparator` ?

alharris-at (approved):

packages/codegen-ui/lib/types/studio-types.ts
dhandsar-aws: Went with `RelationalOperator`

PR #317, opened 2021-12-21
fix: proper use of type from ui builder backend

kpranoto-aws:
*Description of changes:*
JS types specified in the type prop is in lowercase, and all of the values pulled from the backend are string literals. Updating the function to take those into account

### DIFF
--- a/packages/codegen-ui-react/lib/react-component-render-helper.ts
+++ b/packages/codegen-ui-react/lib/react-component-render-helper.ts
@@ -145,34 +145,41 @@ export function buildFixedJsxExpression(prop: FixedStudioComponentProperty): Str
     case 'boolean':
       return factory.createJsxExpression(undefined, value ? factory.createTrue() : factory.createFalse());
     case 'string':
-      switch (type) {
-        case undefined:
-          return factory.createStringLiteral(value as string);
-        case 'String':
-          return factory.createStringLiteral(value as string);
-        case 'Object':
-        case 'Number':
-        case 'Boolean':
-          try {
-            const parsedValue = JSON.parse(value as string);
-
-            if (typeof parsedValue === 'number') {
-              return factory.createJsxExpression(undefined, factory.createNumericLiteral(parsedValue, undefined));
-            }
-            if (typeof parsedValue === 'boolean') {
-              return factory.createJsxExpression(undefined, parsedValue ? factory.createTrue() : factory.createFalse());
-            }
-            // object, array, and null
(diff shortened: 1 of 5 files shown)


yeung-wah:
lgtm, but lets also get a review from @alharris-at/@dpilch on this pr as I am not too familiar with this section of the code

yeung-wah:
also, I think you might want to push this pr to develop instead of main. We generally push changes into develop first, and merge into main when we do the release

packages/codegen-ui-react/lib/react-component-render-helper.ts
```diff
@@ -146,29 +146,32 @@ export function buildFixedJsxExpression(prop: FixedStudioComponentProperty): Str
     case 'string':
       switch (type) {
         case undefined:
-          return factory.createStringLiteral(value as string);
```
dpilch, line 149: This will change the existing behavior when no type is passed. Is this intentional?

```
{
  "value": "{ foo: bar }"
}

<Label prop="{ foo: bar }" />
```

Will change to:

```
{
  "value": "{ foo: bar }"
}

<Label prop={{ foo: bar }} />
```

packages/codegen-ui-react/lib/react-component-render-helper.ts
```diff
@@ -146,29 +146,32 @@ export function buildFixedJsxExpression(prop: FixedStudioComponentProperty): Str
...
+                return factory.createJsxExpression(undefined, factory.createNumericLiteral(parsedValue, undefined));
+              case 'boolean':
+                return factory.createJsxExpression(
+                  undefined,
+                  parsedValue ? factory.createTrue() : factory.createFalse(),
+                );
+              // object, array, and null
+              case 'object':
+                return factory.createJsxExpression(undefined, jsonToLiteral(parsedValue));
+              default:
```
dpilch, line 169: Can you add test cases for these scenarios? Type mismatch case as well.

dpilch (changes requested):

packages/codegen-ui-react/lib/react-component-render-helper.ts
kpranoto-aws: The statement under case `undefined` and case `string` was exactly the same before my changes, I just re-arranged them to reduce code duplication

dpilch: Ah, I understand. You are using fall-through here 👍

dpilch (approved):

PR #320, opened 2021-12-28
fix: append Model to local model imports

dpilch:
In the case where a model shared the name with a prop the generated code would be invalid. Solve this issue by appending `Model` to model import names.

This is only really the first step to a full solution. The customer could still create a prop that is named something like `UserModel` and there would be a name collision. Long term we need to devise a way to prevent name collisions in generated components.

### DIFF
--- a/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react.test.ts.snap
+++ b/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react.test.ts.snap
@@ -148,7 +148,10 @@ export default function Test(props: TestProps): React.ReactElement {
 exports[`amplify render tests collection should render collection with data binding 1`] = `
 "/* eslint-disable */
 import React from \\"react\\";
-import { User, UserPreferences } from \\"../models\\";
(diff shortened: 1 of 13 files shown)


alharris-at:
This change could introduce a regression to customers who are already using the '<ModelName>Model' naming scheme. I think since we have a mitigation in the UI to keep users from entering a colliding name today, we should try and search for a longer-term solution rather than go w/ the bandaid in this case.

Off the top of my head, I think we could look into creating a top-level declaration manager, similar to our imports which gives consistent priority and renames where necessary.

dpilch:
I'm not following the regression case. If they are already using `UserModel` it would import as `UserModelModel`. It would look a little strange but would be functional. But if we are happy with the current mitigation for now I'll go ahead and close this PR.

alharris-at:
If I name my model `PostModel` it'll be renamed as `PostModelModel`, but if I bind a `Post` to the prop named `PostModel` it'll still collide I think (if I'm reading your comment correctly, and thinking about this reasonably). Screenshot of the case is attached.

Screen Shot 2022-01-04 at 10 54 57 AM

PR #327, opened 2022-01-06
feat: add workflow events

dpilch:


dpilch:
Updated to use `eventProperties`. I haven't added the validation to throw an error when a reserved event name is used in `properties`. That can be done as a separate change.

packages/codegen-ui-react/lib/react-component-render-helper.ts
```diff
@@ -404,36 +394,38 @@ export function buildChildElement(prop?: ComponentPropertyValueTypes): JsxChild
...
+  return factory.createJsxAttribute(factory.createIdentifier(name), undefined);
 }
 
 /* Tempory stub function to map from generic event name to React event name. Final implementation will be included in
  * amplify-ui.
  */
-function mapGenericEventToReact(genericEventBinding: string): string {
+export function mapGenericEventToReact(genericEventBinding: string): string {
   switch (genericEventBinding) {
     case 'click':
```
alharris-at, line 425: I feel a bit as though the keys to this map should like the codegen-ui repo, so we can just enforce their names in the yup model.

packages/codegen-ui-react/lib/react-component-with-children-renderer.ts
```diff
@@ -58,9 +59,14 @@ export class ReactComponentWithChildrenRenderer<TPropIn> extends ComponentWithCh
...
-      buildOpeningElementAttributes(value, key),
-    );
+    const attributes = Object.entries(this.component.properties)
+      .map(([key, value]) => buildOpeningElementAttributes(value, key))
+      .concat(
+        Object.entries(this.component.eventProperties || {}).map(([key, value]) =>
+          // pass as bindingProperties to reuse logic
+          buildOpeningElementAttributes({ bindingProperties: { property: value } }, mapGenericEventToReact(key)),
+        ),
+      );
```
alharris-at, line 69: Shared code between the child and standard properties. Not critical in this PR, but we should start looking at how to normalize these two if possible.

alharris-at: Additionally, this logic is slightly opaque, unless you're cached into event handling specifically. May be more readable to generate bindingProperties, generate eventProperties, then return the concat of those two named vars (no logical change).

alharris-at (commented):
Overall LGTM - one style comment is that we should separate renames and just general refactoring from logical changes, this CR would be a bit more readable if it had been split into two.

alharris-at (approved):

PR #347, opened 2022-01-22
feat: remove concept of elementHierarchy from override and variant su…

alharris-at:
…pport, and use component name

*Issue #, if available:*
N/A

*Description of changes:*
No longer use elementHierarchy name for overrides/variants, rely on component name instead. Added validation to ensure component names are unique within a given component schema.


packages/codegen-ui/lib/validation-helper.ts
```diff
@@ -58,20 +58,64 @@ const studioComponentChildSchema: any = yup.object({
...
+    actions: yup.object().nullable(),
+  })
+  // eslint-disable-next-line func-names
+  .test('unique-component-names', 'All component names must be unique', function (value) {
+    const { path, createError } = this;
+    const { duplicateNames } = findCollidingNames(value);
+    if (duplicateNames.size > 0) {
+      return createError({
+        path,
+        message: `Duplicate names are not allowed within a component, found ${JSON.stringify([...duplicateNames])}`,
```
dpilch, line 83: nit: maybe change to `... found duplicates of ['Comp1', 'Comp2']`. At first when I was reading the snapshotted error I was confused because I thought it was saying "DeepComp" and "MyComp" were the same component name.

dpilch (approved):

packages/codegen-ui/lib/validation-helper.ts
alharris-at: Makes sense, I'll update the error verbiage a bit.

PR #349, opened 2022-01-24
feat: add workflow actions

dpilch:
Add workflow actions basic support. I used NavigateAction as the first action, but this does not fully cover the QA for NavigateAction. Integ tests were also not done because the hooks have not been implemented yet. 

To Do in separate changes: 
* unique identifier for action name (wait for new override changes)
* add special case for parameters: model, fields
* add auth attribute support to action parameters (wait for auth attributes merged)


packages/codegen-ui-react/lib/action.ts
```diff
@@ -0,0 +1,137 @@
...
+  const actionName = ActionNameMapping[Action[action as Action]];
+  if (actionName === undefined) {
+    throw new InvalidInputError(`${action} is not a valid action.`);
+  }
+  return actionName;
+}
+
+export function getComponentActions(component: StudioComponent | StudioComponentChild): ActionStudioComponentEvent[] {
+  // TODO: add unique identifier to each action
+  if (component.events) {
```
alharris-at, line 62: You'll miss child-only events as written. You should just treat as two separate conditionals, and create the empty array at the top.

packages/codegen-ui-react/lib/action.ts
```diff
@@ -0,0 +1,137 @@
...
+          ),
+        false,
+      ),
+    );
+  }
+  // TODO: determine what should be used when no parameters
+  return factory.createObjectLiteralExpression([], false);
+}
+
+export function actionParameterToExpression(parameter: StudioComponentProperty): Expression {
```
alharris-at, line 115: We almost certainly support a generic version of this already in the helper, right?

alharris-at: Or are you keeping separate to make it easier to handle the circular dependency thing here?

dpilch: I think we actually don't have a generic version. I wrote this portion around 2 months ago so I can't remember the exact reasoning. It looks like there is repeated logic for isFixedProperty, isBoundProperty, etc. but each adds a different spin to the output. We should refactor all to use one generic version as the base.

* https://github.com/aws-amplify/amplify-codegen-ui/blob/main/packages/codegen-ui-react/lib/react-component-render-helper.ts#L227
* https://github.com/aws-amplify/amplify-codegen-ui/blob/main/packages/codegen-ui-react/lib/react-component-render-helper.ts#L264
* https://github.com/aws-amplify/amplify-codegen-ui/blob/main/packages/codegen-ui-react/lib/react-component-render-helper.ts#L370

I think it's best to do that as a separate change though.

alharris-at: Yep, makes sense to me. I'll create a tracking task.

alharris-at (approved):

PR #358, opened 2022-01-25
feat: add special case for fields and model action parameters

dpilch:
Adds stub for DataStoreCreateItem action but does not fully implement

### DIFF
--- a/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react.test.ts.snap
+++ b/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react.test.ts.snap
@@ -1,5 +1,49 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`amplify render tests actions DataStoreCreate action 1`] = `
+Object {
+  "componentText": "/* eslint-disable */
+import React from \\"react\\";
+import {
+  EscapeHatchProps,
+  getOverrideProps,
+  useDataStoreCreateAction,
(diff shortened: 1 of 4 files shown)


packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react.test.ts.snap
```diff
@@ -1,5 +1,49 @@
...
+export type CreateCustomerButtonProps = React.PropsWithChildren<
+  Partial<ButtonProps> & {
+    overrides?: EscapeHatchProps | undefined | null;
+  }
+>;
+export default function CreateCustomerButton(
+  props: CreateCustomerButtonProps
+): React.ReactElement {
+  const { overrides: overridesProp, ...rest } = props;
+  const overrides = { ...overridesProp };
```
dhandsar-aws, line 24: We can avoid line#24 and use overrides directly? or renaming to `overridesProp` is required?

dhandsar-aws (approved):

dhandsar-aws (changes requested):
Also, tests are failing I'm sure you're aware & already looking into it

packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react.test.ts.snap
dpilch: Existing bug, see other snapshots.

alharris-at (approved):

PR #363, opened 2022-01-26
chore: moving example schemas to top-level, test structure cleanup

alharris-at:
*Issue #, if available:*
N/A

*Description of changes:*
Starting to move our golden file samples to the `codegen-ui` repo instead of react, to make our cross-platform testing more consistent. There were no functional test changes included in this CR.

I avoided moving our snapshots out for now to keep it easier for us to update them as we move forward with GA preparedness, but we should probably lock them once we're GA so any further snapshot changes are very obvious.

I'm thinking longer term we'd have a folder per-golden that's containing:
* "test-name"
  * schema.json
  * react-component.tsx
  * next-platform-component.tsx
  * ...etc


dhandsar-aws (approved):

dpilch (approved):
Nice!

PR #366, opened 2022-01-26
feat: use unique identifier for actions

dpilch:
If you have any suggestions on how the unique identifier should be constructed please leave a comment. I just went with one structure I though was sensible.

I needed to move some functions to a new file to avoid a cyclical dependency between `action.ts` and `react-component-render-helper.ts`.

### DIFF
--- a/packages/codegen-ui-react/lib/react-component-renderer.ts
+++ b/packages/codegen-ui-react/lib/react-component-renderer.ts
@@ -13,13 +13,7 @@
   See the License for the specific language governing permissions and
   limitations under the License.
  */
-import {
-  ComponentRendererBase,
-  StudioNode,
-  StudioComponent,
-  StudioComponentChild,
-  StudioGenericEvent,
-} from '@aws-amplify/codegen-ui';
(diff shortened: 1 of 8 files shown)


packages/codegen-ui-react/lib/react-component-renderer.ts
```diff
@@ -67,7 +57,7 @@ export class ReactComponentRenderer<TPropIn> extends ComponentRendererBase<
       buildOpeningElementProperties(value, key),
     );
     const eventAttributes = Object.entries(this.component.events || {}).map(([key, value]) =>
-      buildOpeningElementEvents(value, mapGenericEventToReact(key as StudioGenericEvent)),
```
dpilch, line 70: Pushed `mapGenerEventToReact` further down so that we would get the same result form `getActionIdentifier`.

alharris-at:
I think to reduce the wordiness of this, we may be able to just generate the action name from the component name and action, since that combination should be unique. The action type probably doesn't need be put in the variable, since it's referenced immediately after in the declaration. That's also easy enough to change later, so this overall LGTM.

alharris-at (approved):

dpilch:
Good idea, I will update to this approach.

PR #371, opened 2022-01-28
feat: add mutations

dpilch:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/codegen-ui-react/lib/react-component-renderer.ts
+++ b/packages/codegen-ui-react/lib/react-component-renderer.ts
@@ -13,7 +13,13 @@
   See the License for the specific language governing permissions and
   limitations under the License.
  */
-import { ComponentRendererBase, StudioNode, StudioComponent, StudioComponentChild } from '@aws-amplify/codegen-ui';
+import {
+  ComponentRendererBase,
+  StudioNode,
+  StudioComponent,
+  StudioComponentChild,
+  StateReference,
+} from '@aws-amplify/codegen-ui';
 import {
   JsxAttributeLike,
   JsxElement,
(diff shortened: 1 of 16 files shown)


packages/codegen-ui/lib/types/studio-types.ts
```diff
@@ -570,6 +543,21 @@ export type DataStoreDeleteItemAction = {
...
+    state: MutationActionSetStateParameter;
+  };
+};
+
+export type MutationActionSetStateParameter = {
+  componentName: string;
+  property: string;
+  set: StudioComponentProperty;
+} & CommonPropertyValues;
+
```
alharris-at, line 558: Can probably remove the CommonPropertyVAlues here, since we made that change in the internal representation as well.

packages/codegen-ui-react/lib/react-component-renderer.ts
```diff
@@ -53,13 +64,27 @@ export class ReactComponentRenderer<TPropIn> extends ComponentRendererBase<
...
     const eventAttributes = Object.entries(this.component.events || {}).map(([key, value]) =>
       buildOpeningElementEvents(value, key, this.component.name),
     );
-    const attributes = propertyAttributes.concat(eventAttributes);
+    const controlEventAttributes = Object.entries(localStateReferences)
+      .filter(([, { addControlEvent }]) => addControlEvent)
+      .map(([key]) =>
+        buildOpeningElementEvents(
+          { bindingEvent: getSetStateName({ componentName: this.component.name || '', property: key }) },
+          'change', // TODO: use component event mapping
```
alharris-at, line 83: Is there a usecase where this isn't going to be `change`? I'm thinking TextField, CheckBoxField, etc. are all basically just listening to the value 'change' event. It doesn't make sense for this to be applied to anything that isn't a field, right?

packages/codegen-ui-react/lib/workflow/mutation.ts
```diff
@@ -0,0 +1,158 @@
...
+
+export function getActionStateParameters(action: ActionStudioComponentEvent): StateStudioComponentProperty[] {
+  if (action.parameters) {
+    return Object.entries(action.parameters)
+      .filter(([key]) => key !== 'model')
+      .map(([key, parameter]) => {
+        if (key === 'fields' || key === 'attributes') {
+          return Object.values(parameter);
+        }
+        return parameter;
```
alharris-at, line 70: I think you can just `flatMap` here instead of additional call to `flat`.

alharris-at (commented):
Overall LGTM - just a few minor questions/comments outside of the existing TODOs.

packages/codegen-ui/lib/types/studio-types.ts
alharris-at: Addressed in the latest commit.

packages/codegen-ui-react/lib/workflow/mutation.ts
alharris-at: Addressed in the latest commit.

alharris-at (approved):

PR #388, opened 2022-02-08
feat: add control event to all state references with change event

dpilch:


packages/codegen-ui-react/lib/primitive.ts
```diff
@@ -108,3 +108,20 @@ export const PrimitiveTypeParameter: Partial<
...
 }
+
+/*
+ * temporary list of Primitives with a change event. Final implementation will pull from amplify UI
+ */
+export const PrimitivesWithChangeEvent: Set<Primitive> = new Set([
+  Primitive.CheckboxField,
+  Primitive.PasswordField,
+  Primitive.PhoneNumberField,
+  Primitive.Radio,
```
alharris-at, line 119: Do we need the underlying radio as well?

packages/test-generator/integration-test-templates/cypress/integration/generate-spec.ts
```diff
@@ -142,7 +142,12 @@ const EXPECTED_SUCCESSFUL_CASES = new Set([
   'DataStoreActions',
   'FormWithState',
   'InternalMutation',
-  'TwoWayBindings',
+  'FullForm',
+  'Listings',
+  'ListingComp',
+  'RoudedImage',
```
alharris-at, line 148: Probably don't need/want to commit this.

alharris-at (commented):
Nice! Just a few minor ones.

packages/test-generator/integration-test-templates/cypress/integration/generate-spec.ts
dpilch: Messed up the rebase. Will fix

packages/codegen-ui-react/lib/primitive.ts
dpilch: No we don't need it, bad copy paste.

packages/codegen-ui-react/lib/workflow/mutation.ts
```diff
@@ -246,17 +247,21 @@ export function filterStateReferencesForComponent(
...
+    mutationReferences: MutationReferences,
+    stateReference: StateReference,
+  ): MutationReferences {
+    const propertyReferences =
+      stateReference.property in mutationReferences ? mutationReferences[stateReference.property] : [];
+    return {
+      ...mutationReferences,
+      [stateReference.property]: propertyReferences.concat([
+        { addControlEvent: PrimitivesWithChangeEvent.has(componentType as Primitive) },
+      ]),
```
alharris-at, line 264: Wouldn't we want the addControlEvent flag here to effectively be either the old logic or this new logic? Is there a usecase where we want to add a control event to a non-form element? Could be not, but it may be safer in the short term to leave the old behavior in until we convince our selves otherwise.

PR #396, opened 2022-02-10
feat: add default initial value for form component state

dpilch:
*Description of changes:*

Set default initial state value for form components. 

List was determined by interacting with TwoWayBinding tests on integration app and adding default initial values for components that gave a warning.

Todo in another change:
If a `defaultValue` is set by the customer and the component becomes controlled the defaultValue needs to be removed because the form components cannot have both `defaultValue` and `value`.

### DIFF
--- a/packages/codegen-ui-react/lib/primitive.ts
+++ b/packages/codegen-ui-react/lib/primitive.ts
@@ -130,9 +130,7 @@ export const PrimitivesWithChangeEvent: Set<Primitive> = new Set([
   Primitive.TextField,
 ]);
 
-export const PrimitiveDefaultPropertyValue: Partial<
-  Record<Primitive, { [property: string]: FixedStudioComponentProperty }>
-> = {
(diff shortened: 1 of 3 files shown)


packages/codegen-ui-react/lib/primitive.ts
```diff
@@ -128,3 +129,29 @@ export const PrimitivesWithChangeEvent: Set<Primitive> = new Set([
   Primitive.SwitchField,
   Primitive.TextField,
 ]);
+
+export const PrimitiveDefaultPropertyValue: Partial<
+  Record<Primitive, { [property: string]: FixedStudioComponentProperty }>
```
alharris-at, line 134: Not sure if it's more succinct or readable, but I think this can be defined as a PrimitiveLevelPropConfig<FixedStudioComponentProperty>

alharris-at (approved):

packages/codegen-ui-react/lib/primitive.ts
dpilch: Oh nice! I didn't notice we had that.

PR #400, opened 2022-02-11
feat: adding method to compute component metadata

alharris-at:
*Issue #, if available:*
N/A

*Description of changes:*
Today, we decorate our component tree w/ metadata as we walk it here and there, but have seen bugs where we're missing detection of auth attributes and state references if they're deeply nested. Creating a new utility method which will replace a lot of this logic, and do a comprehensive walk of the component structure, identifying dependent data models, state references, auth attributes, and component name mapping elements so we can use them without recomputing later.

I've separated the logic here form the actual refactor, to keep this PR focused on the functionality described here.


cwoolum (approved):
LGTM!

PR #413, opened 2022-02-16
chore: sync types with internal

dpilch:
- chore: remove unused types
- chore: sync types with internal


alharris-at (approved):

dpilch:
I changed the `StudioComponent.name` back to optional because we have logic in codegen checking if `name` is set. I will follow up this change with modifications needed to make name required.

PR #428, opened 2022-02-21
feat: poc for useEffect to manage state dependencies

alharris-at:
*Issue #, if available:*
N/A

*Description of changes:*
* Add support for useEffect to update dependent state values once useData or useAuth values have settled.
* Still needs unit tests.
* Still needs improved E2E verification
* Still likely needs to determine how we keep from mutations and state values waffling a bound state element back and forth. (I'd propose we initialize these to `undefined` and then only set the value if the depenedencies are all non-null, and the value itself isn't yet set).


alharris-at (commented):
I can't approve since I have a commit in the PR, but overall looks sane to me.

PR #453, opened 2022-03-09
fix: make complex integ tests agnostic to style order

hein-j:
*Issue #, if available:*

*Description of changes:*
**Problem**
Complex integ tests were failing due to the element styles being applied in a different order than previously.
**Solution**
Because the order of the styles do not matter to functionality, the tests are now made to be indifferent to it.
**Alternative considered**
Q: Why did you not just use e.g. `cy.get(#element).should('have.css', 'color', 'blue')`?
A: I found that it was not reliable because the styles that codegen applies to a given element is sometimes shown as overridden by the parent with this approach.


alharris-at (approved):
Nice!

SwaySway (approved):

hein-j:
PR from forked repo is failing to get credentials necessary for checks. Closing and opening a new PR:
https://github.com/aws-amplify/amplify-codegen-ui/pull/453

PR #456, opened 2022-03-12
chore: refactor out component-with-child-renderer split

aherschel:
This PR removes the mostly duplicate *WithChildRendererBase and *WithChildRenderer classes from the project, since we can just optionally accept a `renderChild` method in the actual renderer, making it pretty straightforward to support components with or without children. This addresses the significant duplicated code between these two classes, and also means we can remove the unnecessary dependency on `@aws-amplicy/ui-react` from `@aws-amplify/codegen-ui` which shouldn't have any dependency on the react-specific packages from UI.

**N.B. I expect the CLI tests to fail in this PR, since I don't have the requisite secrets in my cloned repo to run an `amplify pull` successfully.**


Jshhhh (approved):
LGTM

SwaySway (approved):

PR #469, opened 2022-05-27
chore: add form-related types

hein-j:
*Description of changes:*
As we start codegen work for rendering customer forms, we need to have the types ready.
So far, only the ones that were necessary for the current work-in-progress mapper were added.
You may find, for example, that types for custom validations are not yet there.
They will be added as we understand the process of mapping them better.


SwaySway (approved):
🛳

Jshhhh (approved):
LGTM

PR #488, opened 2022-06-18
feat: add base form renderer and amplify form renderer classes

SwaySway:
*Issue #, if available:*

*Description of changes:*
- add form renderer class
- add amplify form renderer class

### DIFF
--- a/packages/codegen-ui-react/lib/amplify-ui-renderers/form.ts
+++ b/packages/codegen-ui-react/lib/amplify-ui-renderers/form.ts
@@ -1,0 +1,124 @@
+/*
+  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+  Licensed under the Apache License, Version 2.0 (the "License").
+  You may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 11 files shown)


packages/codegen-ui-react/lib/amplify-ui-renderers/form.ts
```diff
@@ -0,0 +1,124 @@
...
+    this.addPropsSpreadAttributes(propsArray);
+
+    return factory.createJsxOpeningElement(
+      factory.createIdentifier(this.component.componentType),
+      undefined,
+      factory.createJsxAttributes(propsArray),
+    );
+  }
+
+  private getFormOnSubmitAttribute(): JsxAttribute {
```
SwaySway, line 72: this is the onSubmit Event for form output would look like
```jsx
<form
      onSubmit={(e) => {
      e.preventDefault();
        const fields = onSubmitBefore();
        // This function will change based on datastore modelname
        myDataStoreFormOnSubmit();
      }}
      {...rest}
      {...getOverrideProps(overrides, "myDataStoreForm")}
    >
```

Jshhhh (approved):

SwaySway:
> General comment: Not sure if we want form-to-component to live inside generate-form-definition? maybe good to move it a level up?

Yup agreed, that's the goal. Will have a that included in a following pr which will be more focused on the following
- form mapping
- form definition
- adding said form def into component metadata

PR #491, opened 2022-06-22
feat: view types and string formatting on view config

kpranoto-aws:
*Changes:*
* Added view types in codegen to be used in Codegen and UI Builder (instead of the type from amplify-studio-common package)
* Had to change the timeFormat string to better match recent required model change

### DIFF
--- a/packages/codegen-ui/lib/__tests__/string-formatter.test.ts
+++ b/packages/codegen-ui/lib/__tests__/string-formatter.test.ts
@@ -40,8 +40,8 @@ describe('string-formatter tests', () => {
       date.setSeconds(23, 222);
 
       expect(formatTime(awsTime, 'locale')).toBe(date.toLocaleTimeString());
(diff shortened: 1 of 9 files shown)


cwoolum (approved):
LGTM

packages/codegen-ui/lib/__tests__/string-formatter.test.ts
```diff
@@ -55,7 +55,7 @@ describe('string-formatter tests', () => {
       const mixedFormatting: DateTimeFormat = {
         dateTimeFormat: {
           dateFormat: 'DD.MM.YYYY',
-          timeFormat: '12hr',
+          timeFormat: 'hours12',
```
SwaySway, line 58: nit: Could we create enums for the date and time formats?

SwaySway (approved):
aside from nit lgtm

PR #492, opened 2022-06-23
fix: optional enum type and nullish handling

kpranoto-aws:
*Changes:*
* nullish string handling for our JS customers
* optional enum type to use for formatting args

### DIFF
--- a/packages/codegen-ui/lib/types/string-format.ts
+++ b/packages/codegen-ui/lib/types/string-format.ts
@@ -14,6 +14,21 @@
   limitations under the License.
  */
 
+export enum DATE {
+  LOCALE = 'locale',
+  MDY = 'MM/DD/YYYY',
+  DMY = 'DD.MM.YYYY',
+  YMD = 'YYYY.MM.DD',
+  MmmDY = 'Mmm DD, YYYY',
+  test = 'test',
+}
+
+export enum TIME {
+  LOCALE = 'locale',
+  HOURS_12 = 'hours12',
+  HOURS_24 = 'hours24',
+}
+
 export type DateFormat = {
   dateFormat: 'locale' | 'MM/DD/YYYY' | 'DD.MM.YYYY' | 'YYYY.MM.DD' | 'Mmm DD, YYYY';
 };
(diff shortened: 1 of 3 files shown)


SwaySway (approved):

hein-j (approved):

packages/codegen-ui/lib/types/string-format.ts
```diff
@@ -14,6 +14,21 @@
   limitations under the License.
  */
 
+export enum DATE {
```
cwoolum, line 17: Can we please not use enums? It adds more work when we have to deserialize. I'd rather we just have a type that represents the available string options.

packages/codegen-ui/lib/utils/string-formatter.ts
```diff
@@ -19,6 +19,10 @@ import { DateFormat, DateTimeFormat, TimeFormat } from '../types';
 const invalidDateStr = 'Invalid Date';
 
 export function formatDate(date: string, format: DateFormat['dateFormat']): string {
+  if (date === undefined || date === null) {
```
cwoolum, line 22: `if(!date)` would cover all cases here.

packages/codegen-ui/lib/utils/string-formatter.ts
```diff
@@ -48,6 +52,10 @@ export function formatDate(date: string, format: DateFormat['dateFormat']): stri
 }
 
 export function formatTime(time: string, format: TimeFormat['timeFormat']): string {
+  if (time === undefined || time === null) {
```
cwoolum, line 55: `if(!time)` would cover all cases here.

packages/codegen-ui/lib/utils/string-formatter.ts
```diff
@@ -79,6 +87,10 @@ export function formatTime(time: string, format: TimeFormat['timeFormat']): stri
 }
 
 export function formatDateTime(dateTimeStr: string, format: DateTimeFormat['dateTimeFormat']): string {
+  if (dateTimeStr === undefined || dateTimeStr === null) {
```
cwoolum, line 90: `if(!dateTimeStr )` would cover all cases here.

PR #499, opened 2022-07-06
feat: support slot binding

hein-j:
*Description of changes:*
- Clean up slot types: Remove `bindingProperties`  from `StudioComponentSlotBinding` and remove `SlotStudioComponentProperty` altogether (use `BoundStudioComponentProperty` instead). Changes to internal packages to follow.
- Add slot of type `React.ReactNode` to the component and bind to children components
- Don't map nested components if `children` property present

### DIFF
--- a/packages/codegen-ui-react/lib/react-component-render-helper.ts
+++ b/packages/codegen-ui-react/lib/react-component-render-helper.ts
@@ -31,6 +31,7 @@ import {
   ActionStudioComponentEvent,
   MutationActionSetStateParameter,
   ComponentMetadata,
+  StudioComponentProperties,
 } from '@aws-amplify/codegen-ui';
 
 import {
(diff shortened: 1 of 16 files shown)


kpranoto-aws:
If children prop is present, I think the intended behavior is to override them with the slot binding. Not 100% sure on this though.

kpranoto-aws (approved):
🚢

packages/codegen-ui-react/lib/react-component-render-helper.ts
```diff
@@ -631,3 +632,7 @@ export function getSetStateName(stateReference: StateStudioComponentProperty): s
   const stateName = getStateName(stateReference);
   return ['set', stateName.charAt(0).toUpperCase() + stateName.slice(1)].join('');
 }
+
+export function hasChildrenProp(componentProperties: StudioComponentProperties): boolean {
+  return Object.keys(componentProperties).includes('children');
```
letsbelopez, line 637: What about .hasOwnProperty('children') instead of creating and looping through arrays?

letsbelopez: actually there's a new version https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn

hein-j: much better, thanks.

SwaySway (approved):

PR #501, opened 2022-07-07
chore: skip field generation for read-only fields

scottyoung:
*Description of changes:*
For forms generated by data model, we do not want to generate fields that are auto-generated read only (such as createdBy or updatedBy).  We also do not want to generate a field for the ID field that is auto generated for every model.

### DIFF
--- a/packages/codegen-ui/lib/__tests__/generate-form-definition/generate-form-definition.test.ts
+++ b/packages/codegen-ui/lib/__tests__/generate-form-definition/generate-form-definition.test.ts
@@ -56,6 +56,52 @@ describe('generateFormDefinition', () => {
     });
   });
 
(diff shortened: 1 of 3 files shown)


hein-j (commented):
Two comments:
1. I think a better approach would be to add everything to `modelFieldsConfigs` but not add the skipped fields to the `elementMatrix`. That way, if the customer does decide to add the fields in through overrides, properties like "label", etc. carry over. 

2. Would you please add a test to the larger `generateFormDefinition` function that makes sure the above happens?

scottyoung:
1.  Added to `modelFieldsConfigs` but will not add the skipped fields to `elementMatrix`
2. Added tests to `generateFormDefinition` to verify 1

hein-j (approved):

cwoolum (approved):
LGTM

PR #527, opened 2022-07-19
chore: add default breakpointhook and styling

felipechiave:
*Issue #, if available:*

*Description of changes:*
On components with variants:
- Added generation of default breakpoint map and integrated into variants
- Added transition style


packages/codegen-ui-react/lib/react-studio-template-renderer.ts
```diff
@@ -594,7 +594,7 @@ export abstract class ReactStudioTemplateRenderer extends StudioTemplateRenderer
       factory.createBindingElement(
         factory.createToken(ts.SyntaxKind.DotDotDotToken),
         undefined,
-        factory.createIdentifier('rest'),
+        factory.createIdentifier(hasVariant ? 'restProp' : 'rest'),
```
Jshhhh, line 597: What is the purpose of using 'restProp' vs 'rest'?

felipechiave: If a component has variants, we want to include a default transition style on the parent component. In a component-with-variants case, using restProp initially lets us define rest with a default transition style later (see line 778 in react-studio-template-renderer.ts)

Jshhhh (approved):
Nice work! LGTM

packages/codegen-ui-react/lib/__tests__/studio-ui-codegen-react.test.ts
```diff
@@ -44,6 +44,12 @@ describe('amplify render tests', () => {
     it('should generate a simple image component', () => {});
 
     it('should generate a simple string component', () => {});
+
+    it('should generate a simple component without variant specific generation', () => {
+      const generatedCode = generateWithAmplifyRenderer('buttonGolden');
+      expect(generatedCode.componentText.indexOf('restProp')).toEqual(-1);
```
letsbelopez, line 50: I think string.contains would be a better method for tests. could read pretty natural like this I think.

expect(generatedCode.componentText.contains('restProp')).toBe(false);

packages/codegen-ui-react/lib/__tests__/studio-ui-codegen-react.test.ts
```diff
@@ -232,6 +238,12 @@ describe('amplify render tests', () => {
       const generatedCode = generateWithAmplifyRenderer('componentWithObjectVariants');
       expect(generatedCode).toMatchSnapshot();
     });
+
+    it('should have variant specific generation', () => {
+      const generatedCode = generateWithAmplifyRenderer('componentWithVariants');
+      expect(generatedCode.componentText.indexOf('restProp')).toBeGreaterThanOrEqual(0);
+      expect(generatedCode.componentText.indexOf('breakpointHook')).toBeGreaterThanOrEqual(0);
```
letsbelopez, line 245: same as above comment

packages/codegen-ui-react/lib/react-studio-template-renderer.ts
```diff
@@ -719,10 +721,103 @@ export abstract class ReactStudioTemplateRenderer extends StudioTemplateRenderer
...
+   *     const breakpointHook = useBreakpointValue({
+   *        base: 'base',
+   *        large: 'large',
+   *        medium: 'medium',
+   *        small: 'small',
+   *        xl: 'xl',
+   *        xxl: 'xxl',
+   *     });
+   */
+  private buildDefaultBreakpointMap() {
```
letsbelopez, line 734: Is it useful to make this into a helper function? A function that creates a hook maybe?

createHookStatement('breakpointHook', 'useBreakpointValue', createObjectLiteralHelper({...base: 'base'}));

letsbelopez (changes requested):
One question and one request in the tests

packages/codegen-ui-react/lib/react-studio-template-renderer.ts
felipechiave: Could be, especially if we want to add additional hooks to components in the future, added.

letsbelopez (approved):

PR #535, opened 2022-07-21
chore: update view types to use string primitives

awinberg-aws:
*Description of changes:*
update types for views to use string primitives instead of the String class.


Jshhhh (approved):
LGTM

hein-j (approved):

PR #542, opened 2022-07-22
feat: add onchange handling for form renderer

Jshhhh:
*Issue #, if available:*

*Description of changes:*
- Adds onchange prop to "*Fields" components  in forms
- Adds state mutation statements and validation for each field

### DIFF
--- a/packages/codegen-ui-react/lib/forms/form-renderer-helper.ts
+++ b/packages/codegen-ui-react/lib/forms/form-renderer-helper.ts
@@ -14,8 +14,15 @@
   limitations under the License.
  */
 
-import { StateStudioComponentProperty, StudioForm, StudioFormActionType } from '@aws-amplify/codegen-ui';
-import { BindingElement, factory, NodeFlags, PropertySignature, SyntaxKind } from 'typescript';
+import {
+  ComponentMetadata,
+  StateStudioComponentProperty,
+  StudioComponent,
+  StudioComponentChild,
+  StudioForm,
+  StudioFormActionType,
+} from '@aws-amplify/codegen-ui';
+import { BindingElement, Expression, factory, NodeFlags, PropertySignature, SyntaxKind } from 'typescript';
 import { ImportCollection, ImportSource, ImportValue } from '../imports';
 import { getStateName, getSetStateName } from '../react-component-render-helper';
 import { addSchemaToArguments, getActionHookImportValue, getActionIdentifier } from '../workflow';
@@ -29,6 +36,10 @@ export const FieldStateVariable = (componentName: string): StateStudioComponentP
   componentName,
   property: 'fields',
 });
+
+function capitalizeFirstLetter(val: string) {
+  return val.charAt(0).toUpperCase() + val.slice(1);
+}
 
 /**
  * - formFields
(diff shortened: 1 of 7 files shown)


packages/codegen-ui-react/lib/forms/form-renderer-helper.ts
```diff
@@ -280,9 +284,160 @@ export const buildFormPropNode = (form: StudioForm) => {
...
+                            factory.createStringLiteral(fieldName),
+                          ),
+                          undefined,
+                          [factory.createIdentifier('value')],
+                        ),
+                      ),
+                      factory.createToken(SyntaxKind.ColonToken),
+                      factory.createCallExpression(factory.createIdentifier('validateField'), undefined, [
+                        factory.createIdentifier('value'),
+                        factory.createIdentifier(`${fieldName}-validation-rules`),
```
Jshhhh, line 383: Creating a distinct string here to replace with validation rules

packages/codegen-ui-react/lib/forms/react-form-renderer.ts
```diff
@@ -123,7 +124,14 @@ export abstract class ReactFormTemplateRenderer extends StudioTemplateRenderer<
 
     const wrappedFunction = this.renderFunctionWrapper(this.component.name, variableStatements, jsx, false);
 
-    const result = printer.printNode(EmitHint.Unspecified, wrappedFunction, file);
+    let result = printer.printNode(EmitHint.Unspecified, wrappedFunction, file);
+    if (this.componentMetadata.formMetadata?.onValidationFields) {
+      Object.entries(this.componentMetadata.formMetadata?.onValidationFields).forEach(
+        ([fieldName, validationRules]) => {
+          result = result.replace(`${fieldName}-validation-rules`, JSON.stringify(validationRules));
```
Jshhhh, line 131: Replaces distinct validation rules string with stringified validationRules array. I did this because validationRules array is already in the shape needed for the validation function

cwoolum (approved):
LGTM

packages/codegen-ui/lib/generate-form-definition/helpers/form-field.ts
```diff
@@ -52,6 +52,7 @@ export function getFormDefinitionInputElement(
       formDefinitionElement = {
         componentType: 'TextField',
         props: {
+          name: config.inputType?.name || baseConfig?.inputType?.name || FORM_DEFINITION_DEFAULTS.field.inputType.name,
```
Jshhhh, line 55: Added name prop for the field component to have reference to its field in data model during form codegen

hein-j: You can find the name as the key for the element. In the definition. You can pass that in if that's not getting passed in now. "name" is actually a prop on some elements with a separate meaning. I think we should reserve "props" for things that are actually going to be mapped as props on the ui-react component.

Jshhhh: Thanks I was worried about that. Ill have to find a way to connect the form definition to the mapped studio component

hein-j (approved):

packages/codegen-ui-react/lib/forms/form-renderer-helper.ts
```diff
@@ -280,9 +284,159 @@ export const buildFormPropNode = (form: StudioForm) => {
   return factory.createTypeLiteralNode(propSignatures);
 };
 
+export const buildStateMutationStatement = (name: string, defaultValue: any) => {
```
SwaySway, line 287: can we type this to an expression?
```suggestion
export const buildStateMutationStatement = (name: string, defaultValue: Expression) => {
```

Jshhhh: Yup. Will update

PR #548, opened 2022-07-27
feat: support for cta labels

scottyoung:
*Issue #, if available:*

*Description of changes:*
Updates FormDefinition and StudioForm to accept StudioFormButtons to customize CTA label names, visibility, and row position.


packages/codegen-ui/lib/generate-form-definition/form-to-component.ts
```diff
@@ -106,7 +106,10 @@ export const ctaButtonConfig = (): StudioComponentChild => {
         name: 'CancelButton',
         properties: {
           label: {
-            value: 'Cancel',
+            value:
+              formDefinition.buttons?.cancel?.visible && formDefinition.buttons?.cancel?.labelOverride
+                ? formDefinition.buttons?.cancel?.labelOverride
+                : 'Cancel',
```
SwaySway, line 112: Can we have section above which renders the value with formDef and handles the default value as well?

scottyoung: added resolveCtaLabels section to handle this

SwaySway (approved):

packages/codegen-ui/lib/generate-form-definition/form-to-component.ts
```diff
@@ -88,7 +88,30 @@ export const fieldComponentMapper = (name: string, formDefinition: FormDefinitio
...
+
+  const clearLabel =
+    formDefinition.buttons?.clear?.visible && formDefinition.buttons?.clear?.labelOverride
+      ? formDefinition.buttons?.clear?.labelOverride
+      : 'Clear';
+
+  const submitLabel =
+    formDefinition.buttons?.submit?.visible && formDefinition.buttons?.submit?.labelOverride
+      ? formDefinition.buttons?.submit?.labelOverride
+      : 'Submit';
```
letsbelopez, line 107: Do we need to add localization support?

packages/codegen-ui/lib/generate-form-definition/generate-form-definition.ts
```diff
@@ -72,5 +77,9 @@ export function generateFormDefinition({
 
   formDefinition.form.layoutStyle = mapStyles(form.style);
 
+  if (form.ctaConfig) {
```
letsbelopez, line 80: if this is an empty object it will be true. is that intentional?

packages/codegen-ui/lib/types/form/form-cta.ts
```diff
@@ -0,0 +1,42 @@
...
+type StudioFormButton = VisibleStudioFormButton | HiddenStudioFormButton;
+
+/**
+ * Configuration for each of the specified CTA's
+ */
+export type StudioFormButtons = {
+  /**
+   * The position of the CTA's in the form when rendered
+   */
+  position?: 'Top' | 'Bottom' | 'TopAndBottom';
```
letsbelopez, line 35: is this the standard casing we're using?

packages/codegen-ui/lib/types/form/index.ts
```diff
@@ -52,6 +53,8 @@ export type StudioForm = {
   sectionalElements: { [elementName: string]: SectionalElement };
 
   style: StudioFormStyle;
+
+  ctaConfig?: StudioFormButtons;
```
letsbelopez, line 57: Can we add e2e tests for this new prop?

letsbelopez, line 57: Can we change the name of this type to match the property name better? Maybe StudioFormCTAConfig

packages/codegen-ui/lib/generate-form-definition/form-to-component.ts
```diff
@@ -88,7 +88,30 @@ export const fieldComponentMapper = (name: string, formDefinition: FormDefinitio
...
-export const ctaButtonConfig = (): StudioComponentChild => {
+const resolveCtaLabels = (
+  formDefinition: FormDefinition,
+): { cancelLabel: string; clearLabel: string; submitLabel: string } => {
+  const cancelLabel =
+    formDefinition.buttons?.cancel?.visible && formDefinition.buttons?.cancel?.labelOverride
+      ? formDefinition.buttons?.cancel?.labelOverride
+      : 'Cancel';
+
+  const clearLabel =
```
hein-j, line 99: Please move the defaults to the default object in the form definition mapper. `helpers/defaults.ts`

packages/codegen-ui/lib/types/form/form-cta.ts
```diff
@@ -0,0 +1,42 @@
...
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+ */
+type VisibleStudioFormButton = {
+  visible: true;
+
+  labelOverride?: string;
```
hein-j, line 19: Nit - instead of `labelOverride`, how about matching the field configurations and just using `label`?

packages/codegen-ui/lib/types/form/form-cta.ts
```diff
@@ -0,0 +1,42 @@
...
+  visible: false;
+};
+
+type StudioFormButton = VisibleStudioFormButton | HiddenStudioFormButton;
+
+/**
+ * Configuration for each of the specified CTA's
+ */
+export type StudioFormButtons = {
+  /**
```
hein-j, line 32: This shape for CTAs does not support changing the position of the buttons within the CTA row. This functionality is descoped, but I think it should be supported by the shape now to avoid breaking changes to model later

packages/codegen-ui/lib/generate-form-definition/form-to-component.ts
hein-j: This way, the UI will get the defaults too and we won't have a discrepancy. 
General note - you will probably want to do some mapping in the definition mapper (to apply the defaults, and also to figure out the position of the buttons in relation to one another)

packages/codegen-ui/lib/generate-form-definition/form-to-component.ts
SwaySway: I don't think this something we can add in locally as whatever client app would use these components would also need to have localization support. In the UI console we can do it by passing down the override?

scottyoung: Yes, the UI code that will accompany these changes will pass in a default override of localized strings for 'Submit', 'Cancel' and 'Clear'

packages/codegen-ui/lib/generate-form-definition/generate-form-definition.ts
scottyoung: Yes empty objects will be allowed.  Added a test to confirm expectations.

packages/codegen-ui/lib/types/form/form-cta.ts
scottyoung: updated to use camelCase

packages/codegen-ui/lib/types/form/index.ts
scottyoung: updated

packages/codegen-ui/lib/generate-form-definition/form-to-component.ts
scottyoung: moved defaults to _helpers/defaults.ts_

packages/codegen-ui/lib/types/form/form-cta.ts
scottyoung: updated to _label_

packages/codegen-ui/lib/types/form/form-cta.ts
scottyoung: I hadn't considered order within the row itself.  I've updated _StudioFormButton_ with an optional parameter for _position_, since I don't think we've finalized how those positions would look.

hein-j: position should be the same type as the position for the sectional elements and input elements.

packages/codegen-ui/lib/generate-form-definition/form-to-component.ts
hein-j: That is definitely one part. What remains is applying the defaults within the definition mapper. You can look at how inputs use the defaults, for example. You will want to create a helper that--among other things--applies these defaults so buttons always have labels, both in the UI and in the customer project. 

It would be best if the shape of the buttons config produced by the definition mapper is different from the buttons config passed in... a shape that is more digestible by the form-to-components helpers. Again, you can see how inputs are mapped for example

scottyoung: Thanks for the explanation, it's clearer to me now how defaults work with definition mapper.  I've added _mapButtons_ to a new helper file to set defaults

packages/codegen-ui/lib/types/form/form-cta.ts
scottyoung: Updated to use _StudioFieldPosition_

letsbelopez (dismissed):

packages/codegen-ui/lib/generate-form-definition/form-to-component.ts
```diff
@@ -88,7 +89,19 @@ export const fieldComponentMapper = (name: string, formDefinition: FormDefinitio
   return parentGrid(`${name}Grid`, formDefinition.form.layoutStyle, fieldChildren);
 };
 
-export const ctaButtonConfig = (): StudioComponentChild => {
+const resolveCtaLabels = (
+  formDefinition: FormDefinition,
+): { cancelLabel: string; clearLabel: string; submitLabel: string } => {
+  const cancelLabel = formDefinition.buttons.cancel?.label || FORM_DEFINITION_DEFAULTS.ctaConfig.cancel.label;
+  const clearLabel = formDefinition.buttons.clear?.label || FORM_DEFINITION_DEFAULTS.ctaConfig.clear.label;
```
hein-j, line 96: You should not need to use the defaults here, since you are applying them in the definition mapper

packages/codegen-ui/lib/generate-form-definition/form-to-component.ts
```diff
@@ -106,7 +119,7 @@ export const ctaButtonConfig = (): StudioComponentChild => {
         name: 'CancelButton',
         properties: {
           label: {
-            value: 'Cancel',
+            value: cancelLabel,
```
hein-j, line 122: just noticed, also, that you would want to pass in variants to get those differences in style

packages/codegen-ui/lib/types/form/form-definition.ts
```diff
@@ -24,7 +25,7 @@ export type FormDefinition = {
     layoutStyle: StudioFormStyle;
   };
   elements: { [element: string]: FormDefinitionElement };
-  buttons: { [key: string]: string };
+  buttons: StudioFormCTAConfig;
```
hein-j, line 28: if you make this shape different from StudioFormCTAConfig, you can make it a lot more digestible by the renderer. For example, mapping the positions of the buttons in relation to each other, making label a required field, adding variants

SwaySway (approved):

hein-j:
Hey, by the way of elaborating on[ this comment,](https://github.com/aws-amplify/amplify-codegen-ui/pull/548#discussion_r938211369) I'll propose a shape:

```
// you can keep this one in form-definition-element
type FormDefinitionButtonElement = {
  componentType: 'Button';
  props: {
    variant?: string;
    children: string;
  }

}

// this goes on like formDefinition.buttons
type buttons = {
  buttonConfigs: {
    submit?: FormDefinitionButtonElement; // optional because they can be hidden
    cancel?: FormDefinitionButtonElement;
    clear?: FormDefinitionElement;
  }
  position: 'top' | 'bottom' | 'topAndBottom';
  buttonMatrix: ('submit' | 'cancel' | 'clear')[][] // e.g. [['clear']['cancel', 'submit']] to represent what we currently have in the UI
}
```

SwaySway (approved):

packages/codegen-ui/lib/utils/form-to-component/map-form-definition-to-component.ts
```diff
@@ -98,55 +98,84 @@ const ctaButtonConfig = (): StudioComponentChild => {
...
+    });
+  }
+
+  CTAComponent.children?.push(rightAlignCTA);
+
+  return CTAComponent;
+};
+
+const renderChildren = (name: string, formDefinition: FormDefinition) => {
+  switch (formDefinition.buttons.position) {
```
hein-j, line 165: What do you think about creating a file for button-related helpers?

scottyoung: This is a good suggestion, I don't want to bloat this file - do you want it added to this PR or in a follow up?

packages/codegen-ui/lib/utils/form-to-component/map-form-definition-to-component.ts
```diff
@@ -98,55 +98,84 @@ const ctaButtonConfig = (): StudioComponentChild => {
...
+    case 'top':
+      return [ctaButtonMapper(formDefinition), fieldComponentMapper(name, formDefinition)];
```
hein-j, line 167: I feel like this logic can be broken into a helper function. Something like
```
function addCTA(formChildren: [], position, buttons: StudioComponentInstance) {
const nextFormChildren = [...formChildren]
   switch(position) {
  case 'top':
   nextFormChildren.splice(0, 0, buttons)
   

}

return nextChildren
}
```

packages/codegen-ui/lib/utils/form-to-component/map-form-definition-to-component.ts
```diff
@@ -98,55 +98,84 @@ const ctaButtonConfig = (): StudioComponentChild => {
...
+      },
+    });
+  }
+
+  if (formDefinition.buttons.buttonConfigs.submit) {
+    rightAlignCTA.children?.push({
+      componentType: 'Button',
+      name: 'SubmitButton',
+      properties: {
+        children: {
```
hein-j, line 146: The advantage of having `props` on the form definition element is that you don't have to hand map like this. If you look at how the other elements are mapped, you can see. Also, componentType and name can be mapped in a DRY way too. For name, you can do a simple string conversion or mapping.

For type and variation, you can also map this from the form definition element. ('variant' should be 'variation'. My bad)

packages/codegen-ui/lib/utils/form-to-component/map-form-definition-to-component.ts
```diff
@@ -98,55 +98,84 @@ const ctaButtonConfig = (): StudioComponentChild => {
...
+    children: [],
+  };
+
+  if (formDefinition.buttons.buttonConfigs.cancel) {
+    CTAComponent.children?.push({
+      componentType: 'Button',
+      name: 'CancelButton',
+      properties: {
+        children: {
+          value: formDefinition.buttons.buttonConfigs.cancel.props.children,
```
hein-j, line 110: you could create a helper that maps buttons instead of doing it one by one. Something like

```
mapButton(buttonElement, buttonName // this would be the key you get when Object.entries) {
  return {componentType: element.componentType,
name: transformIntoButtonname(buttonName),
properties: mapFormElementProps(element) // you can map all the props like type and variation (not variant like you have now. my bad) in the form definition mapper all you gotta do is just map them on
```

hein-j (dismissed):

packages/codegen-ui/lib/utils/form-to-component/helpers/map-cta-buttons.ts
```diff
@@ -0,0 +1,116 @@
...
+  const updatedFormChildren = formChildren || [];
+
+  switch (position) {
+    case 'top':
+      updatedFormChildren.splice(0, 0, buttons);
+      break;
+    case 'bottom':
+      updatedFormChildren.splice(formChildren.length, 0, buttons);
+      break;
+    case 'topAndBottom':
```
letsbelopez, line 106: Where are these position values defined? Is it an enum or union type somewhere? if so, can we update it to a standardized enum format value? "top_and_bottom"

letsbelopez (changes requested):

letsbelopez (approved):

hein-j (approved):

PR #553, opened 2022-08-03
feat: table definition for renderer

kpranoto-aws:
*Description of changes:*
* Table definition to be used by table renderer
* Some types are fixed to use primitive instead of instanced 

### DIFF
--- a/packages/codegen-ui/lib/generate-view-definition/helpers/order-filter-columns.ts
+++ b/packages/codegen-ui/lib/generate-view-definition/helpers/order-filter-columns.ts
@@ -1,0 +1,130 @@
+/*
+  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+  Licensed under the Apache License, Version 2.0 (the "License").
+  You may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


packages/codegen-ui/lib/generate-view-definition/helpers/order-filter-columns.ts
```diff
@@ -0,0 +1,130 @@
...
+    current = current.next;
+  }
+
+  return ordered;
+};
+
+export const orderAndFilterVisibleColumns = (columns: ColumnsMap, fields?: DataStoreModelField[]): ColumnInfo[] => {
+  let ordered: ColumnInfo[] = [];
+
+  if (fields) {
```
cwoolum, line 105: Need to check that fields has length here also or line 110 will fail.. Please add a test for this also.

kpranoto-aws: Added size check and test for not throwing error

cwoolum (approved):
LGTM

SwaySway (approved):
🚢

PR #556, opened 2022-08-03
Make form labels use sentence case

cwoolum:
*Issue #, if available:*

*Description of changes:*
This changes the labels to be sentence case which was an original requirement but was missed in the AC for the previous tasks.

* Also fixes enums to actually use Title case instead of Sentence Case
* Cleans up `mapModelFieldsConfigs` to no longer take in a mutable object.


packages/codegen-ui/lib/generate-form-definition/generate-form-definition.ts
```diff
@@ -39,15 +39,19 @@ export function generateFormDefinition({
     elementMatrix: [],
   };
 
-  const modelFieldsConfigs: ModelFieldsConfigs = {};
+  let modelFieldsConfigs: ModelFieldsConfigs = {};
```
lyonsbp, line 42: Why can't this just be declared as `const` on line 50?

lyonsbp (approved):

packages/codegen-ui/lib/generate-form-definition/generate-form-definition.ts
cwoolum: Line 50 is inside of a conditional so it wouldn't be in scope at line 57.

hein-j:
The requirements for Enums are that they are actually sentence cased. It's just we've all been misnaming it "title case". 

From the task (let me know if I can send you the link):

> By default, the enum values will have "Title case with spacing"  (convert "-" and "_" to space) and are shown as "placeholders".

hein-j (approved):

PR #566, opened 2022-08-04
chore: separate out and export logic to map model field to field config

hein-j:
*Issue #, if available:*

*Description of changes:*
separate out and export logic to map model field to field config so that it can be used in Amplify Studio.
Needed in order to take advantage of the case conversion for label and enum mappings for previews.


cwoolum (approved):
LGTM

SwaySway (approved):
⛵️

PR #569, opened 2022-08-04
chore: only generate breakpoint mapping on breakpoint variants

felipechiave:
*Issue #, if available:*

*Description of changes:*
Components should only have default breakpoint mapping and transition styles when there is a 'breakpoint' variant

### DIFF
--- a/packages/codegen-ui-react/lib/react-studio-template-renderer.ts
+++ b/packages/codegen-ui-react/lib/react-studio-template-renderer.ts
@@ -21,6 +21,7 @@ import {
   isEventPropertyBinding,
   isStudioComponentWithCollectionProperties,
   isStudioComponentWithVariants,
+  isStudioComponentWithBreakpoints,
   StudioComponent,
   StudioComponentChild,
   StudioComponentPredicate,
@@ -581,6 +582,7 @@ export abstract class ReactStudioTemplateRenderer extends StudioTemplateRenderer
 
     // remove overrides from rest of props
     const hasVariant = isStudioComponentWithVariants(component);
+    const hasBreakpoint = isStudioComponentWithBreakpoints(component);
     elements.push(
       factory.createBindingElement(
         undefined,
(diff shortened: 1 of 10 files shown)


packages/codegen-ui-react/lib/react-studio-template-renderer.ts
```diff
@@ -595,7 +597,7 @@ export abstract class ReactStudioTemplateRenderer extends StudioTemplateRenderer
       factory.createBindingElement(
         factory.createToken(ts.SyntaxKind.DotDotDotToken),
         undefined,
-        factory.createIdentifier(hasVariant ? 'restProp' : 'rest'),
+        factory.createIdentifier(hasBreakpoint ? 'restProp' : 'rest'),
```
SwaySway, line 600: Are restProps only used for breakpoints?

packages/codegen-ui-react/lib/react-studio-template-renderer.ts
```diff
@@ -837,6 +843,42 @@ export abstract class ReactStudioTemplateRenderer extends StudioTemplateRenderer
     );
   }
 
+  /**
```
SwaySway, line 846: not sure I'm seeing the difference between this and `buildOverridesFromVariantsAndRestProp`

packages/codegen-ui-react/lib/react-studio-template-renderer.ts
felipechiave: Yes, restProp is used to include a transition style onto components but it should only be added on components with a 'breakpoint' variant.

packages/codegen-ui-react/lib/react-studio-template-renderer.ts
felipechiave: `buildOverridesFromVariantsAndRestProp` is used when a 'breakpoint' variant exists on a component as it includes the default breakpoint mapping as an argument into `getOverridesFromVariants`. This function builds the call without that mapping for components that do not have a 'breakpoint' variant.

SwaySway (approved):

packages/codegen-ui-react/lib/react-studio-template-renderer.ts
felipechiave: Merged functions while conditioning on `hasBreakpoints`

hein-j (approved):

PR #574, opened 2022-08-09
feat: initial commit for view (table) renderer

kpranoto-aws:
*Description of changes:*
Initial commit so we can build from this

### DIFF
--- a/packages/codegen-ui-react/lib/views/react-view-renderer.ts
+++ b/packages/codegen-ui-react/lib/views/react-view-renderer.ts
@@ -1,0 +1,92 @@
+/*
+  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+  Licensed under the Apache License, Version 2.0 (the "License").
+  You may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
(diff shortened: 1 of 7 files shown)


packages/codegen-ui-react/lib/views/react-view-renderer.ts
```diff
@@ -0,0 +1,105 @@
...
+    if (component.dataSource.model) {
+      const dataModel = dataSchema?.models[component.dataSource.model]?.fields ?? {};
+      fields = Object.entries(dataModel).map(([key, value]) => ({
+        name: key,
+        type: value.dataType,
+        isReadOnly: value.readOnly,
+        isArray: value.isArray,
+        isRequired: value.required,
+      }));
+    }
```
SwaySway, line 76: can we refactor the table def to accept the generic data schema? This logic looks specific to table def mapping

kpranoto-aws: Yep, that makes sense. I will put this mapping in the table def then

SwaySway (approved):

SwaySway (approved):

letsbelopez (approved):

PR #576, opened 2022-08-10
feat: add override types to form elements

Jshhhh:
*Issue #, if available:*

*Description of changes:*
- Add override types for Form elements
- 
```
export declare type FormCustomCreateOverridesProps = {
    FormCustomCreateGrid: GridProps;
    RowGrid0: GridProps;
    firstName: TextFieldProps;
    RowGrid1: GridProps;
    loggedInColor: TextFieldProps;
} & EscapeHatchProps;
export declare type FormCustomCreateProps = React.PropsWithChildren<{
    overrides?: FormCustomCreateOverridesProps | undefined | null;
} & {
    onSubmit: (fields: Record<string, string>) => void;
    onCancel?: () => void;
}>;
```

### DIFF
--- a/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
+++ b/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
@@ -121,8 +121,16 @@ export default function customDataForm(props) {
 exports[`amplify form renderer tests custom form tests should render a custom backed form 2`] = `
 "import * as React from \\"react\\";
 import { EscapeHatchProps } from \\"@aws-amplify/ui-react/internal\\";
+import { GridProps, TextFieldProps } from \\"@aws-amplify/ui-react\\";
+export declare type customDataFormOverridesProps = {
+    customDataFormGrid: GridProps;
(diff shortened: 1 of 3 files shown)


packages/codegen-ui-react/lib/imports/import-mapping.ts
```diff
@@ -16,6 +16,7 @@
 export enum ImportSource {
   REACT = 'react',
   UI_REACT = '@aws-amplify/ui-react',
+  UI_REACT_TYPES = '@aws-amplify/ui-react/dist/types',
```
SwaySway, line 19: do we need this path to import the types?

Jshhhh: Thanks, we dont need to. Will update

hein-j:
Probably out of scope here, but this makes me think -- does form names get properly cased/ sanitized/ or vetted so react doesn't throw?

hein-j (approved):

Jshhhh:
> Probably out of scope here, but this makes me think -- does form names get properly cased/ sanitized/ or vetted so react doesn't throw?

Yeah, we dont sanitize form Names. I'll wrap it in quotes in the type

SwaySway (approved):

PR #577, opened 2022-08-10
feat: render options for SelectField and RadioGroupField

hein-j:
*Description of changes:*
Transform `valueMappings` into `option` children for `SelectField` and `Radio` children for `RadioGroupField`. 
Not in scope - handling `bindingProperties` of `valueMappings`, which have to do with relationships

### DIFF
--- a/packages/codegen-ui/lib/utils/form-to-component/helpers/map-element-children.ts
+++ b/packages/codegen-ui/lib/utils/form-to-component/helpers/map-element-children.ts
@@ -1,0 +1,60 @@
+/*
+  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+  Licensed under the Apache License, Version 2.0 (the "License").
+  You may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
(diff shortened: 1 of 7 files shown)


packages/codegen-ui/lib/utils/form-to-component/helpers/map-element-children.ts
```diff
@@ -0,0 +1,60 @@
...
+import {
+  FormDefinitionElement,
+  FormDefinitionRadioGroupFieldElement,
+  FormDefinitionSelectFieldElement,
+  StudioComponentChild,
+} from '../../../types';
+
+type MapElementChildrenReturnValue = { children: StudioComponentChild[] };
+
+function mapOptions(
```
SwaySway, line 25: should this just return the component children list?

hein-j: I made it an object on purpose because when we handle relationship options later, we will be handling `bindingProperties` and probably returning something like `modelsToImport` alongside the `children`.

SwaySway (approved):
🚢

Jshhhh (approved):

PR #578, opened 2022-08-11
feat: table renderer

kpranoto-aws:
*Description of changes:*
* Basic table renderer

### DIFF
--- a/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-views.test.ts.snap
+++ b/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-views.test.ts.snap
@@ -1,0 +1,47 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`amplify view renderer tests should generate a table element 1`] = `
+"<Table>
+  {!disableHeaders && (
+    <TableHead>
+      <TableRow>
+        <TableCell as=\\"th\\">hireDate</TableCell>
+        <TableCell as=\\"th\\">comments</TableCell>
+        <TableCell as=\\"th\\">createdAt</TableCell>
+        <TableCell as=\\"th\\">updatedAt</TableCell>
+      </TableRow>
+    </TableHead>
+  )}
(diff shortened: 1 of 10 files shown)


packages/codegen-ui-react/lib/react-table-renderer.ts
```diff
@@ -0,0 +1,274 @@
...
+                  factory.createIdentifier('item'),
+                  factory.createIdentifier(columnId),
+                ),
+              ],
+            ),
+            factory.createToken(SyntaxKind.ColonToken),
+            factory.createPropertyAccessExpression(
+              factory.createIdentifier('item'),
+              factory.createIdentifier(columnId),
+            ),
```
SwaySway, line 196: Don't we want to pass the item, the format type, and the formatting spec?

kpranoto-aws: Need to rethink this one since customers can provide their own formatting by passing override, we also have configured formatting from CP, and no formatting (literal value).

SwaySway: Ah okay let's add changes as a follow up

SwaySway (approved):

packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-views.test.ts.snap
```diff
@@ -0,0 +1,47 @@
...
+        <TableCell as=\\"th\\">createdAt</TableCell>
+        <TableCell as=\\"th\\">updatedAt</TableCell>
+      </TableRow>
+    </TableHead>
+  )}
+  <TableBody>
+    {items.map((item, index) => (
+      <TableRow
+        onClick={
+          highlightOnHover && onRowClick ? () => onRowClick(item, index) : null
```
letsbelopez, line 19: fyi, I'm not 100% the highlightOnHover needs to be enabled for onRowClick, we may need to change this later.

letsbelopez (approved):

PR #579, opened 2022-08-11
fix: codegenning formatter in table jsx

kpranoto-aws:
*Description of changes:*
* using formatter in codegenned table jsx

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #584, opened 2022-08-12
feat: formatter init

letsbelopez:
*Issue #, if available:*

*Description of changes:*
Adding formatting util generation


packages/codegen-ui-react/lib/utils/format.ts
```diff
@@ -0,0 +1,1125 @@
...
+export function format(value: string, formatterInput: FormatInputType) {
+  switch (formatterInput.type) {
+    case 'DateFormat':
+      return formatDate(value, formatterInput.format);
+    case 'DateTimeFormat':
+      return formatDateTime(value, formatterInput.format);
+    case 'TimeFormat':
+      return formatTime(value, formatterInput.format);
+    default:
+      return undefined;
```
SwaySway, line 158: if we call the formatter on values that are undefined for format input - i think we should just return the value? or is the format util only being called on fields which need a format?

packages/codegen-ui-react/lib/__tests__/utils/__snapshots__/format.test.ts.snap
```diff
@@ -0,0 +1,103 @@
...
+    const dateAndTime = dateTime.toISOString().split(\\"T\\");
+    const date = formatDate(dateAndTime[0], format.dateFormat);
+    const time = formatTime(dateAndTime[1], format.timeFormat);
+    return \`\${date} - \${time}\`;
+}
+export function format(value: string, formatterInput: FormatInputType) {
+    switch (formatterInput.type) {
+        case \\"DateFormat\\": return formatDate(value, formatterInput.format);
+        case \\"DateTimeFormat\\": return formatDateTime(value, formatterInput.format);
+        case \\"TimeFormat\\": return formatTime(value, formatterInput.format);
```
kpranoto-aws, line 100: We want to handle undefined for `value` and `formatterInput`. Both should return the value as is

packages/codegen-ui-react/lib/utils/format.ts
letsbelopez: this is the return never TS issue

letsbelopez: but eslint requires a default return

packages/codegen-ui-react/lib/__tests__/utils/__snapshots__/format.test.ts.snap
```diff
@@ -0,0 +1,103 @@
...
+    }
+    if (format === \\"locale\\") {
+        return dateTime.toLocaleString();
+    }
+    const dateAndTime = dateTime.toISOString().split(\\"T\\");
+    const date = formatDate(dateAndTime[0], format.dateFormat);
+    const time = formatTime(dateAndTime[1], format.timeFormat);
+    return \`\${date} - \${time}\`;
+}
+export function format(value: string, formatterInput: FormatInputType) {
```
kpranoto-aws, line 96: The function name should be `formatter`

packages/codegen-ui-react/lib/utils/format.ts
kpranoto-aws: The TS issue was because the last statement was inside the switch statement. We want to check for undefined for both value and formatterInput before it reaches the switch statement. If any of them is undefined, return the unformatted value

packages/codegen-ui-react/lib/__tests__/utils/__snapshots__/format.test.ts.snap
letsbelopez: you like formatter? it sounds like a noun instead of a verb

packages/codegen-ui-react/lib/utils/format.ts
letsbelopez: I think returning the value inside the switch statement should be fine.

packages/codegen-ui-react/lib/__tests__/utils/__snapshots__/format.test.ts.snap
letsbelopez: we don't check for undefined before calling the function? Usually I check the inputs before.

packages/codegen-ui-react/lib/__tests__/utils/__snapshots__/format.test.ts.snap
kpranoto-aws: format might cause name collision with customer's `format` which is a map they provide to put their own custom formatters

packages/codegen-ui-react/lib/__tests__/utils/__snapshots__/format.test.ts.snap
kpranoto-aws: Yeah but we can't assume anything since it's codegenned for customers and they can provide their own input

letsbelopez: oh it's already being handled in the functions being called, formatDate, formatDateTime etc

packages/codegen-ui-react/lib/__tests__/utils/__snapshots__/format.test.ts.snap
letsbelopez: that should be called formatOverride to keep with the whole override naming convention in components

packages/codegen-ui-react/lib/__tests__/utils/__snapshots__/format.test.ts.snap
kpranoto-aws: `formatterInput` can be undefined so that has to be handled before the switch statement otherwise it would cause an error. undefined values are passed in formatterInput when customers don't have their custom `format` map and they did not configure any formatting in UI Builder

letsbelopez: In that case there should be a default formatter. dates for example should always use a default format

letsbelopez: if we're passing undefined formatting configs, then I think there's something wrong with our code

letsbelopez: [This](https://github.com/aws-amplify/amplify-codegen-ui/pull/579/files#diff-13b04baba7c792bc0221d51e887f6ea6ecf8eb0349078e529b62cf06af628435R34) shouldn't blindly be calling formatter when we don't need it. Is it possible to make it smarter? If it's undefined, don't call formatter and just use the value.

kpranoto-aws: Yeah that's what I originally suggested. I don't mind going that route

PR #597, opened 2022-08-18
chore: change form validation

hein-j:
*Description of changes:*
- extend existing validation instead of replacing it if custom validation function
- factor out the validation tasks so that it can be reused on submit later (currently only used on change)
- fix logic for getting if form is valid.
- remove empty error messages on mount

### DIFF
--- a/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
+++ b/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
@@ -3,11 +3,11 @@
 exports[`amplify form renderer tests custom form tests should render a custom backed form 1`] = `
 "/* eslint-disable */
 import * as React from \\"react\\";
-import { validateField } from \\"@aws-amplify/codegen-ui-react\\";
 import {
   getOverrideProps,
   useStateMutationAction,
 } from \\"@aws-amplify/ui-react/internal\\";
+import { validateField } from \\"@aws-amplify/codegen-ui-react\\";
 import {
   Button,
   Flex,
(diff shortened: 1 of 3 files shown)


packages/codegen-ui-react/lib/forms/form-renderer-helper.ts
```diff
@@ -451,7 +492,7 @@ export const addFormAttributes = (
   return attributes;
 };
 
-export const buildOnChangeStatement = (
+export const XXbuildOnChafdafdngeStatement = (
```
Jshhhh, line 495: I presume this is slightly unintentional? haha

hein-j: d'oh! thanks, deleted the whole block

packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
```diff
@@ -25,24 +25,23 @@ export default function customDataForm(props) {
...
   const [modelFields, setModelFields] = useStateMutationAction({});
   const [formValid, setFormValid] = useStateMutationAction(true);
+  const [errors, setErrors] = useStateMutationAction({});
+  const validations = {
+    name: [],
+    email: [],
+    city: [],
+    category: [],
+  };
+  async function runValidationTasks(fieldName, value) {
```
SwaySway, line 37: nit: can we use an arrow function?

SwaySway (commented):
since the errors is starting with an object can we use optional chaining for instances like below?
```
errorMessage={errors.name.errorMessage}
hasError={errors.name.hasError}
```

packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
hein-j: done

hein-j:
> since the errors is starting with an object can we use optional chaining for instances like below?
> 
> ```
> errorMessage={errors.name.errorMessage}
> hasError={errors.name.hasError}
> ```

good call. done

SwaySway (approved):

Jshhhh (approved):
LGTM

PR #605, opened 2022-08-26
chore: update type for view

SwaySway:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/codegen-ui/lib/types/view/view.ts
+++ b/packages/codegen-ui/lib/types/view/view.ts
@@ -18,14 +18,16 @@ import { ViewStyle } from './style';
 import { ColumnsMap } from './table';
 
 export interface StudioView {
-  appId?: string;
-  dataSource: ViewDataTypeConfig;
-  environmentName?: string;
-  id: string;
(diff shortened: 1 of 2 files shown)


packages/codegen-ui/lib/types/view/view.ts
```diff
@@ -18,14 +18,16 @@ import { ViewStyle } from './style';
 import { ColumnsMap } from './table';
 
 export interface StudioView {
-  appId?: string;
```
hein-j, line 21: One thing to note is that this interface is used to instantiate StudioView model in UI.

hein-j: I don't think it'll break UI but just a note

hein-j (approved):

packages/codegen-ui/lib/types/view/view.ts
hein-j: Maybe sync w/ Kevin

SwaySway:
closing this pr as the original implementation has changed for view types

PR #606, opened 2022-08-29
fix: remove @aws-amplify/codegen-ui-react as import source

SwaySway:
By removing this import generated forms can now use the utils file for
validateField use. This shouldn't be an issue for UI as that does not
render the import source for the files and we can decide the import
value in the UI code.

*Issue #, if available:*

*Description of changes:*


Jshhhh (approved):
LGTM

letsbelopez (approved):

PR #607, opened 2022-08-29
fix: remove pascal case from theme names

SwaySway:
fix: remove pascal case from themes as the default theme name is "studioTheme"

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/codegen-ui/lib/validation-helper.ts
+++ b/packages/codegen-ui/lib/validation-helper.ts
@@ -17,6 +17,10 @@ import * as yup from 'yup';
 import { InvalidInputError } from './errors';
 import { StudioGenericEvent, StudioSchema } from './types';
 
+const alphaNumString = () => {
+  return yup.string().matches(/^[a-zA-Z0-9]*$/, { message: 'Expected an alphanumeric string' });
+};
+
 const pascalCaseAlphaNumString = () => {
   return yup
     .string()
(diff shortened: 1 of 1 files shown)


Jshhhh (approved):

packages/codegen-ui/lib/validation-helper.ts
```diff
@@ -173,7 +177,7 @@ const studioThemeValuesSchema: any = yup.object({
 });
 
 const studioThemeSchema = yup.object({
-  name: pascalCaseAlphaNumString().required(),
+  name: alphaNumString().required(),
```
letsbelopez, line 180: maybe a better name is camelCaseAlphaNumString

letsbelopez (approved):

packages/codegen-ui/lib/validation-helper.ts
```diff
@@ -17,6 +17,10 @@ import * as yup from 'yup';
 import { InvalidInputError } from './errors';
 import { StudioGenericEvent, StudioSchema } from './types';
 
+const alphaNumString = () => {
+  return yup.string().matches(/^[a-zA-Z0-9]*$/, { message: 'Expected an alphanumeric string' });
```
letsbelopez, line 21: should the error message say Expected a camel case alphanumeric string?

letsbelopez (approved):

packages/codegen-ui/lib/validation-helper.ts
letsbelopez: can't start with a digit, need to fix in a follow up

packages/codegen-ui/lib/validation-helper.ts
SwaySway: will fix as a follow up pr needs to ref this pr convo

PR #613, opened 2022-08-30
Form arrays

Jshhhh:
*Issue #, if available:*

*Description of changes:*
- Adds configuration for Array type input fields

Next PR:
- ArrayField component


packages/codegen-ui/lib/generate-form-definition/helpers/model-fields-configs.ts
```diff
@@ -65,6 +69,7 @@ export function getFieldConfigFromModelField({
       readOnly: field.readOnly,
       name: fieldName,
       value: 'true',
+      isArray: field.isArray,
```
hein-j, line 72: What is the reason for adding this? I think I'm missing where it's used.

Jshhhh: This will be used when the util function is generated. We will need to export our own ArrayField component that will exist in the utils file

hein-j (approved):

SwaySway (approved):

PR #615, opened 2022-08-30
chore: add phone validation, change required validation, and fix regex

hein-j:
*Description of changes:*
- Add validation for `PhoneNumberField` (bare bones according to AWSPhone scalar spec)
- Add validation for required fields.
- Do not run validations other than `Required` if input is empty
- Fix validation regex that were not generating correctly in cx project

### DIFF
--- a/packages/codegen-ui-react/lib/__tests__/forms/validation.test.ts
+++ b/packages/codegen-ui-react/lib/__tests__/forms/validation.test.ts
@@ -20,7 +20,6 @@ describe('validateField tests', () => {
   it('should validate REQUIRED type', () => {
     expect(validateField('123', [{ type: ValidationTypes.REQUIRED, validationMessage: '' }])).toEqual({
       hasError: false,
-      errorMessage: 'The value is required',
     });
     expect(validateField('', [{ type: ValidationTypes.REQUIRED, validationMessage: '' }])).toEqual({
       hasError: true,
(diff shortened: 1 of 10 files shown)


packages/codegen-ui-react/lib/__tests__/forms/validation.test.ts
```diff
@@ -297,4 +296,26 @@ describe('validateField tests', () => {
...
+    expect(validateField('293 849 3029', [{ type: ValidationTypes.PHONE, validationMessage: 'test' }])).toEqual({
+      hasError: false,
+      errorMessage: 'test',
+    });
+
+    expect(validateField('293-849-3029', [{ type: ValidationTypes.PHONE, validationMessage: 'test' }])).toEqual({
+      hasError: false,
+      errorMessage: 'test',
+    });
+  });
```
Jshhhh, line 320: what if the number has 1 hyphen and 1 space? 123-123 1234

packages/codegen-ui-react/lib/utils/forms/validation.ts
```diff
@@ -123,12 +131,27 @@ export const validateField = (
...
+  phone: '/^[0-9 -]+$/',
+};
```
Jshhhh, line 153: Well this is annoying. :(

packages/codegen-ui-react/lib/__tests__/forms/validation.test.ts
hein-j: Looks like DataStore will allow it.

https://github.com/aws-amplify/amplify-js/blob/46ee5dd91c61f49bad4da8286b2f97c737d96631/packages/datastore/src/util.ts#L675-L677

Rev'd to use this exact regex and added a test case for your point.

Jshhhh: Thanks!

Jshhhh (approved):

letsbelopez (approved):

PR #621, opened 2022-09-02
fix: do not assume related model or field is in datastore schema

hein-j:
*Description of changes:*
The `getGenericFromDataStore` was throwing when there is an associated field spelled out, but it does not actually exist in the schema.
This PR handles when the field is missing. For good measure, it also handles if an associated model were to be missing, although that has not been raised as an issue.


SwaySway (approved):
🛳

awinberg-aws (approved):

PR #631, opened 2022-09-08
fix: add default values for update form

Jshhhh:
*Issue #, if available:*

*Description of changes:*
- Adds default values for Update form
- update each field state on datastore query
- fix: remove label for ArrayField component

### DIFF
--- a/packages/codegen-ui/lib/types/form/form-metadata.ts
+++ b/packages/codegen-ui/lib/types/form/form-metadata.ts
@@ -16,6 +16,11 @@
 import { DataFieldDataType } from '../data';
 import { FieldValidationConfiguration } from './form-validation';
 
+/**
+ * Form Action type definition
+ */
+export type StudioFormActionType = 'create' | 'update';
+
 export type FieldConfigMetadata = {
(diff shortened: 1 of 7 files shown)


packages/codegen-ui/lib/types/form/form-metadata.ts
```diff
@@ -16,6 +16,11 @@
 import { DataFieldDataType } from '../data';
 import { FieldValidationConfiguration } from './form-validation';
 
+/**
+ * Form Action type definition
+ */
+export type StudioFormActionType = 'create' | 'update';
```
SwaySway, line 22: nit: i think we might want to keep this with the main form type

SwaySway (approved):
aside nit -> ⛴

packages/codegen-ui/lib/types/form/form-metadata.ts
Jshhhh: Yeah, I moved it to avoid circular imports. The main form type should be moved out of index.ts

hein-j (approved):

PR #632, opened 2022-09-09
fix: test all configured validations for field

awinberg-aws:
## Problem
`validateField` reports no error for fields with multiple validations configured if it passes the first check, even if it would fail any of the remaining conditions.

## Solution
Refactor `validateField` to only return the result of a rule check if a validation error was found, allowing other configured rules to be checked.

## Additional Changes
n/a

### Automated tests
- [x] Automated tests added
- [x] Automated tests updated

### DIFF
--- a/packages/codegen-ui-react/lib/__tests__/forms/validation.test.ts
+++ b/packages/codegen-ui-react/lib/__tests__/forms/validation.test.ts
@@ -13,7 +13,7 @@
   See the License for the specific language governing permissions and
   limitations under the License.
  */
-import { ValidationTypes } from '@aws-amplify/codegen-ui/lib/types/form/form-validation';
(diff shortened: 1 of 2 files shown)


Jshhhh (approved):

packages/codegen-ui-react/lib/__tests__/forms/validation.test.ts
```diff
@@ -305,22 +294,37 @@ describe('validateField tests', () => {
...
-      errorMessage: 'test',
     });
 
     expect(validateField('293 849-3029', [{ type: ValidationTypes.PHONE, validationMessage: 'test' }])).toEqual({
       hasError: false,
-      errorMessage: 'test',
+    });
+  });
+
+  it('should test value against all configured validations', () => {
```
Jshhhh, line 312: Nice catch!

hein-j (approved):
I like how we're not returning error messages anymore if there's no error

PR #640, opened 2022-09-13
feat: add onChange override function

SwaySway:
*Issue #, if available:*

*Description of changes:*
- add `onChange` on form types
- call `onChange` if the prop is passed down
- slight refactor on how variable statements are created

### DIFF
--- a/packages/codegen-ui-react/lib/__tests__/react-forms/__snapshots__/form-renderer-helper.test.ts.snap
+++ b/packages/codegen-ui-react/lib/__tests__/react-forms/__snapshots__/form-renderer-helper.test.ts.snap
@@ -6,6 +6,7 @@ exports[`form-render utils should generate before & complete types if datastore 
     onSuccess?: (fields: Record<string, unknown>) => void;
     onError?: (fields: Record<string, unknown>, errorMessage: string) => void;
     onCancel?: () => void;
+    onChange?: (fields: Record<string, unknown>) => Record<string, unknown>;
     onValidate?: mySampleFormInputValues;
 }"
 `;
@@ -14,6 +15,7 @@ exports[`form-render utils should generate regular onsubmit if dataSourceType is
 "{
     onSubmit: (fields: Record<string, unknown>) => void;
     onCancel?: () => void;
+    onChange?: (fields: Record<string, unknown>) => Record<string, unknown>;
     onValidate?: myCustomFormInputValues;
 }"
 `;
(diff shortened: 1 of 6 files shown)


packages/codegen-ui-react/lib/forms/form-renderer-helper.ts
```diff
@@ -480,12 +506,67 @@ export function buildOnBlurStatement(fieldName: string) {
...
+            ],
+            NodeFlags.Const,
+          ),
+        ),
+        factory.createExpressionStatement(
+          factory.createBinaryExpression(
+            factory.createIdentifier('value'),
+            factory.createToken(SyntaxKind.EqualsToken),
+            factory.createBinaryExpression(
+              buildAccessChain(['result', ...fieldName.split('.')]),
```
SwaySway, line 551: this is to build the access chain to support nested json (example in snapshot diff and below)
```js
value = result?.bio?.favoriteAnimal ?? value;
```

packages/codegen-ui-react/lib/__tests__/react-forms/__snapshots__/form-renderer-helper.test.ts.snap
```diff
@@ -6,6 +6,7 @@ exports[`form-render utils should generate before & complete types if datastore
     onSuccess?: (fields: Record<string, unknown>) => void;
     onError?: (fields: Record<string, unknown>, errorMessage: string) => void;
     onCancel?: () => void;
+    onChange?: (fields: Record<string, unknown>) => Record<string, unknown>;
```
cwoolum, line 9: Shouldn't the type be specifically related to the fields in the form? We should be able to infer all of those.

cwoolum (commented):
One question

packages/codegen-ui-react/lib/__tests__/react-forms/__snapshots__/form-renderer-helper.test.ts.snap
SwaySway: I can create a follow up pr to create the types for datastore-backed forms, though for custom forms it's bit more unknown as we'd need a larger task to track those data types.
cc: @hein-j

cwoolum: Even for JSON forms though, we still know the types that the different primitives return so we can still provide a map of fields to those types.

cwoolum (approved):
Please follow up with the proper types in a later CR

hein-j (approved):

PR #643, opened 2022-09-14
chore: refine types and onChange handlers

hein-j:
*Description of changes:*
- Change handlers and types to account for differences across component types.
Verified by creating a form. E2E test would be useful to prevent regressions here. That will follow in another PR

### DIFF
--- a/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
+++ b/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
@@ -12,6 +12,7 @@ import {
   Radio,
   RadioGroupField,
   SelectField,
+  StepperField,
   TextField,
 } from \\"@aws-amplify/ui-react\\";
 export default function CustomDataForm(props) {
@@ -21,12 +22,14 @@ export default function CustomDataForm(props) {
   const [email, setEmail] = React.useState(undefined);
   const [city, setCity] = React.useState(undefined);
   const [category, setCategory] = React.useState(undefined);
+  const [pages, setPages] = React.useState(undefined);
   const [errors, setErrors] = React.useState({});
   const validations = {
     name: [{ type: \\"Required\\" }],
     email: [{ type: \\"Required\\" }],
     city: [],
     category: [],
+    pages: [],
   };
   const runValidationTasks = async (fieldName, value) => {
     let validationResponse = validateField(value, validations[fieldName]);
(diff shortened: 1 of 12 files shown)


packages/codegen-ui-react/lib/forms/event-targets.ts
```diff
@@ -17,57 +17,101 @@ import { DataFieldDataType } from '@aws-amplify/codegen-ui';
...
+  let defaultIdentifier: string | BindingName =
+    fieldTypeToExpressionMap[fieldType]?.identifier ?? expressionMap.destructuredValue;
   switch (dataType) {
-    // const value = Number(new Date(e.target.value));
     case 'AWSTimestamp':
+      // value = Number(new Date(e.target.value));
+      defaultIdentifier = expressionMap.value;
       expression = factory.createCallExpression(factory.createIdentifier('Number'), undefined, [
         factory.createNewExpression(factory.createIdentifier('Date'), undefined, [
           factory.createPropertyAccessExpression(expression, factory.createIdentifier('value')),
```
SwaySway, line 81: nit: we can re-use the value here as well from the expressionMap

packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
```diff
@@ -1897,7 +1933,7 @@ export declare type InputGalleryCreateFormInputValues<useBase extends boolean =
     attend?: UseBaseOrValidationType<useBase, boolean>;
     maybeSlide?: UseBaseOrValidationType<useBase, boolean>;
     maybeCheck?: UseBaseOrValidationType<useBase, boolean>;
-    arrayTypeField?: UseBaseOrValidationType<useBase, string>;
+    arrayTypeField?: UseBaseOrValidationType<useBase, string[]>;
```
SwaySway, line 1936: i think we run validation on each element in the array as opposed to passing the whole list

packages/codegen-ui-react/lib/forms/event-targets.ts
hein-j: nit you say?? this prevented a regression. thanks!

packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
hein-j: Ok, I reverted the change to array types for now. If I find it needs changes, I'll PR that separately.

cwoolum (approved):
LGTM

letsbelopez (approved):

SwaySway (approved):
lgtm 🚀

PR #645, opened 2022-09-15
fix: include minimum breakpoints required

SwaySway:
*Issue #, if available:*

*Description of changes:*
- adds the minimum amount of breakpoint configs required for the hook to work
- sorts the breakpoints according to size as the base will use the smallest size if not specified


cwoolum (approved):
LGTM

scottyoung (approved):

PR #650, opened 2022-09-16
fix: onChange overrides to use onChange value and add useEffect initial data for update custom forms

SwaySway:
*Issue #, if available:*

*Description of changes:*
- add initialData prop to pass for on update custom forms
  - set the state in a useEffect similar to on update for datastore backed forms
- fix override onchange to pass the value with the model field object into the onChange hook
- move `buildFormPropNode` to type-helper
- remove defaultValue from controlled components as the value will take its place


letsbelopez (approved):

Jshhhh (approved):

cwoolum (approved):
L G T M

PR #665, opened 2022-09-21
feat: add handling for isArray in form field model

awinberg-aws:
## Problem
Studio model now has an isArray property for form fields which isn't being handled when generating form definitions

## Solution
Update type and helper methods to include support for isArray.

### DIFF
--- a/packages/codegen-ui-react/lib/utils/forms/array-field-component.ts
+++ b/packages/codegen-ui-react/lib/utils/forms/array-field-component.ts
@@ -21,7 +21,6 @@ import {
   getCurrentValueName,
   getSetNameIdentifier,
 } from '../../forms/form-state';
-import { lowerCaseFirst } from '../../helpers';
 
 export const generateArrayFieldComponent = () => {
   const iconPath = 'M10 10l5.09-5.09L10 10l5.09 5.09L10 10zm0 0L4.91 4.91 10 10l-5.09 5.09L10 10z';
@@ -1031,7 +1030,7 @@ export const renderArrayFieldComponent = (fieldName: string, fieldLabel: string,
         ),
         factory.createJsxAttribute(
           factory.createIdentifier('inputFieldRef'),
-          factory.createJsxExpression(undefined, factory.createIdentifier(`${lowerCaseFirst(fieldName)}Ref`)),
+          factory.createJsxExpression(undefined, factory.createIdentifier(`${fieldName}Ref`)),
         ),
       ]),
     ),
--- a/packages/codegen-ui/lib/__tests__/generate-form-definition/generate-form-definition.test.ts
+++ b/packages/codegen-ui/lib/__tests__/generate-form-definition/generate-form-definition.test.ts
@@ -33,16 +33,75 @@ describe('generateFormDefinition', () => {
         dataSourceType: 'DataStore',
         enums: {},
         nonModels: {},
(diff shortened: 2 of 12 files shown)


packages/codegen-ui-react/lib/utils/forms/array-field-component.ts
```diff
@@ -1031,7 +1030,7 @@ export const renderArrayFieldComponent = (fieldName: string, fieldLabel: string,
         ),
         factory.createJsxAttribute(
           factory.createIdentifier('inputFieldRef'),
-          factory.createJsxExpression(undefined, factory.createIdentifier(`${lowerCaseFirst(fieldName)}Ref`)),
+          factory.createJsxExpression(undefined, factory.createIdentifier(`${fieldName}Ref`)),
```
awinberg-aws, line 1033: Resolves undefined error. The ref being referenced here is not declared using lowerCaseFirst.

https://github.com/aws-amplify/amplify-codegen-ui/blob/main/packages/codegen-ui-react/lib/forms/react-form-renderer.ts#L435

packages/codegen-ui/lib/generate-form-definition/helpers/form-field.ts
```diff
@@ -339,6 +339,7 @@ export function getFormDefinitionInputElement(
 
   formDefinitionElement.validations = mergedValidations;
   formDefinitionElement.dataType = config?.dataType || baseConfig?.dataType;
+  formDefinitionElement.isArray = config.inputType?.isArray || baseConfig?.inputType?.isArray;
```
cwoolum, line 343: I think we want the DataStore `isArray` config to take priority. If a user has a DataStore field that is just a scalar, they shouldn't be able to override that with an array but that might be a @renebrandel question.

cwoolum: Similarly for the reverse, if the field is a scalar and the user configures it as an array, does the use have to map it correctly in code?

cwoolum (commented):
Have some questions for you and Rene

packages/codegen-ui/lib/__tests__/generate-form-definition/generate-form-definition.test.ts
```diff
@@ -398,6 +413,7 @@ it('should add read-only fields if it has overrides', () => {
     name: {
       componentType: 'TextField',
       dataType: 'String',
+      isArray: false,
```
hein-j, line 416: I think you have to let the metadata get its `isArray` field from the form def element now, or it won't get picked up during codegen. The change is in `form-component-metadata.ts`

packages/codegen-ui/lib/generate-form-definition/helpers/form-field.ts
awinberg-aws: new version prioritizes value in baseConfig if present

packages/codegen-ui/lib/__tests__/generate-form-definition/generate-form-definition.test.ts
awinberg-aws: discussed offline. change was not needed based on previous implementation, but a new version was pushed which no longer relies on referencing dataSchema.

cwoolum (approved):
LGTM

hein-j (approved):

PR #670, opened 2022-09-22
chore: differentiate reset for datastore update form

hein-j:
*Description of changes:*
- For all update forms, change clear button label to "Reset"
- For DataStore update forms, reset the value to that queried from DataStore instead of just clearing
- Dynamically map cta button matrix for use in the UI

Other notes:
- Custom update forms has an outstanding bug tracked in a different ticket
- Changed identifier of the sub-cta flex because there's a task later where buttons can move around


lyonsbp (approved):

cwoolum (approved):
LGTM

PR #672, opened 2022-09-23
fix: timestamp and number fields

SwaySway:
*Issue #, if available:*

*Description of changes:*
- update timestamp fields to first check for valid date before converting to number

```jsx
onChange={(e) => {
        const date = new Date(e.target.value);
        if(!(date instanceof Date && !isNaN(date))) {
          setErrors((error) => ({ ...errors, [fieldName]: 'The value must be a valid date' }))
          return;
        }
        let value = Number(date);
        // rest of code
      }}
```

- added a NaN check for number fields which get parsed
```jsx
onChange={(e) => {
        let value = parseInt(e.target.value);
          if (NaN(value)) {
            setErrors((errors) => ({
              ...errors,
              num: \\"Value must be a valid number\\",
            }));
            return;
        // rest of code
      }}
```


cwoolum (approved):
LGTM

packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
```diff
@@ -3358,7 +3372,15 @@ export default function InputGalleryCreateForm(props) {
         isReadOnly={false}
         type=\\"datetime-local\\"
         onChange={(e) => {
-          let value = Number(new Date(e.target.value));
+          const date = new Date(e.target.value);
+          if (!(date instanceof Date && !isNaN(date))) {
```
cwoolum, line 3376: Actually, if the field isn't required, empty actually should be a valid value for the input . I think that's the case for all inputs we are casting.

cwoolum (changes requested):

packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
SwaySway: in regards to datetime-local fields the onChange is not executed for incomplete dates or empty dates

Jshhhh (approved):

cwoolum (approved):

PR #675, opened 2022-09-26
fix: timestamp formatting in defaultValue

SwaySway:
*Issue #, if available:*
*Description of changes:*
- handle defaultValues date associated fields
  - for `AWSTimestamp` this will convert the epoch seconds/milliseconds to datetime-local input
  - for `AWSDateTime` will remove the 'Z' character so the date is laid out correctly in the input field

note: seconds/milliseconds are not editable
### format for epoch in seconds
Screen Shot 2022-09-26 at 1 48 03 PM

### format for epoch in milliseconds
Screen Shot 2022-09-26 at 1 49 05 PM

### DIFF
--- a/packages/codegen-ui-react/lib/forms/component-helper.ts
+++ b/packages/codegen-ui-react/lib/forms/component-helper.ts
@@ -1,0 +1,65 @@
+/*
+  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+  Licensed under the Apache License, Version 2.0 (the "License").
+  You may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+ */
+
+import { FieldConfigMetadata } from '@aws-amplify/codegen-ui';
+import { BinaryExpression, factory, SyntaxKind } from 'typescript';
+
+export const ControlledComponents = ['StepperField', 'SliderField', 'SelectField', 'ToggleButton', 'SwitchField'];
+
+/**
+ * given the component returns true if the component is a controlled component
+ *
+ * @param componentType
+ * @returns
+ */
+export const isControlledComponent = (componentType: string): boolean => ControlledComponents.includes(componentType);
+
+export const defaultValueAttributeMap: Record<string, (stateName: string) => BinaryExpression> = {
+  AWSTimestamp: (stateName: string) =>
+    factory.createBinaryExpression(
+      factory.createIdentifier(stateName),
+      factory.createToken(SyntaxKind.AmpersandAmpersandToken),
+      factory.createCallExpression(factory.createIdentifier('convertToLocal'), undefined, [
+        factory.createCallExpression(factory.createIdentifier('convertTimeStampToDate'), undefined, [
+          factory.createIdentifier(stateName),
+        ]),
+      ]),
+    ),
+  AWSDateTime: (stateName: string) =>
+    factory.createBinaryExpression(
+      factory.createIdentifier(stateName),
+      factory.createToken(SyntaxKind.AmpersandAmpersandToken),
+      factory.createCallExpression(factory.createIdentifier('convertToLocal'), undefined, [
+        factory.createNewExpression(factory.createIdentifier('Date'), undefined, [factory.createIdentifier(stateName)]),
+      ]),
+    ),
+};
+
+/**
+ *
+ * @param stateName name of the state name to render for the default value
+ * @param { dataType } the dataType
+ * @returns
+ */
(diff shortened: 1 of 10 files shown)


packages/codegen-ui-react/lib/utils/forms/value-mappers.ts
```diff
@@ -0,0 +1,166 @@
...
+
+/*
+given a timestamp number it will convert it to datetime local for forms
+first does a check to see if the timestamp is in seconds or milliseconds
+*/
+export const getDateTimeLocal = (ts: number) => {
+  if (Math.abs(Date.now() - ts) < Math.abs(Date.now() - ts * 1000)) {
+    return new Date(ts).toISOString().replace('Z', '');
+  }
+  return new Date(ts * 1000).toISOString().replace('Z', '');
```
awinberg-aws, line 27: this isn't a "local" conversion so much as it's just stripping off the UTC marker. Why was this necessary, and are there any data integrity concerns here?

SwaySway: This change is added so the date timestamp values being passed in are viewable in an update form.
This shouldn't change the date value itself as its only being used in defaultValue.
Not opposed to not supporting the defaultValue is there a concern where we don't want this value mapper?

hein-j (dismissed):

packages/codegen-ui-react/lib/forms/component-helper.ts
```diff
@@ -0,0 +1,67 @@
...
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+ */
+
+import { FieldConfigMetadata } from '@aws-amplify/codegen-ui';
+import { factory, SyntaxKind } from 'typescript';
+
+export const ControlledComponents = ['StepperField', 'SliderField', 'SelectField'];
```
letsbelopez, line 20: SwitchField and ToggleButton are most likely controlled as well. What's the impact of not including all controlled components?

packages/codegen-ui-react/lib/forms/component-helper.ts
```diff
@@ -0,0 +1,67 @@
...
+          factory.createIdentifier(stateName),
+          factory.createIdentifier('replace'),
+        ),
+        undefined,
+        [factory.createStringLiteral('Z'), factory.createStringLiteral('')],
+      ),
+    );
+  }
+
+  return factory.createJsxAttribute(factory.createIdentifier('defaultValue'), expression);
```
letsbelopez, line 66: I think the defaultValue attribute name is going to change per component type. I think we need a map of componenttype to default value attribute.

For example,

packages/codegen-ui-react/lib/utils/forms/value-mappers.ts
SwaySway: synced offline - reformatting the date so the local time is preserved only when presenting the value

packages/codegen-ui-react/lib/forms/component-helper.ts
SwaySway: adding a map in the next revision

awinberg-aws (approved):

packages/codegen-ui-react/lib/forms/component-helper.ts
SwaySway: the impact is minimal though the idea is to not render default values for these controlled components as the value is coming from the state - will add the other two

packages/codegen-ui-react/lib/utils/forms/value-mappers.ts
```diff
@@ -0,0 +1,360 @@
...
+    year: 'numeric',
+    month: '2-digit',
+    day: '2-digit',
+    hour: '2-digit',
+    minute: '2-digit',
+    second: '2-digit',
+    fractionalSecondDigits: 3,
+  });
+  const parts = df.formatToParts(date).map((v) => v.value);
+  return `${parts[4]}-${parts[0]}-${parts[2]}T${parts[6]}:${parts[8]}:${parts[10]}.${parts[12]}`;
```
awinberg-aws, line 50: This will not work for PM, `df.formatToParts(new Date('Mon Sep 26 2022 20:55:58 GMT-0700 (Pacific Daylight Time)'))` gives `parts[8] === "08"` instead of `"20"`

awinberg-aws (changes requested):

packages/codegen-ui-react/lib/utils/forms/value-mappers.ts
awinberg-aws: Adding `hour12: false` to the config object above will address the AM/PM issue but another thing to be aware of is that the order of parts also changes depending on the locale used by DateTimeFormat. Not sure if `'default'` is meant to vary by the user's actual locale or not but that could work against us for this implementation.

SwaySway: new rev
- changing to set hour12 to false
- for the times we can just index by the key as we know thats a constant for the response from date time format

cwoolum (approved):

PR #679, opened 2022-09-27
fix: accept 0 for number fields

letsbelopez:
…y string

*Issue #, if available:*

https://app.asana.com/0/1202966106210868/1203050023493661/f

*Description of changes:*

I updated the addItem code to allow for the number 0

I updated the default values for string fields to be an empty string. This prevents react components from going from uncontrolled to controlled components.

### DIFF
--- a/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
+++ b/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
@@ -730,7 +730,12 @@ function ArrayField({
     setSelectedBadgeIndex(undefined);
   };
   const addItem = async () => {
-    if ((!!currentFieldValue || currentFieldValue === false) && !hasError) {
+    if (
+      (currentFieldValue !== undefined ||
+        currentFieldValue !== null ||
+        currentFieldValue !== \\"\\") &&
+      !hasError
+    ) {
       const newItems = [...items];
       if (selectedBadgeIndex !== undefined) {
         newItems[selectedBadgeIndex] = currentFieldValue;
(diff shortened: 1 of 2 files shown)


packages/codegen-ui-react/lib/forms/form-state.ts
```diff
@@ -120,7 +120,12 @@ export const getDefaultValueExpression = (
   if (componentType in componentTypeToDefaultValueMap) {
     return componentTypeToDefaultValueMap[componentType];
   }
-  return factory.createIdentifier('undefined');
+  /**
+   * All other types are strings.
+   * Use an empty string here instead of undefined because undefined will cause the component to go from uncontrolled
+   * to controlled. A component should be controlled or uncontrolled for the lifetime of the component.
+   */
+  return factory.createStringLiteral('');
```
letsbelopez, line 128: Previous update called this function to get defaultValues. But I had to update this to return an empty string for all other types.

packages/codegen-ui-react/lib/utils/forms/array-field-component.ts
```diff
@@ -252,18 +252,24 @@ export const generateArrayFieldComponent = () => {
...
+                              factory.createNull(),
                             ),
                           ),
                           factory.createToken(SyntaxKind.BarBarToken),
                           factory.createBinaryExpression(
                             factory.createIdentifier('currentFieldValue'),
-                            factory.createToken(SyntaxKind.EqualsEqualsEqualsToken),
-                            factory.createFalse(),
+                            factory.createToken(SyntaxKind.ExclamationEqualsEqualsToken),
+                            factory.createStringLiteral(''),
```
letsbelopez, line 272: This if statement is updated to allow number and float array fields to accept a number `0` to be added to the array

packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
```diff
@@ -730,7 +730,12 @@ function ArrayField({
     setSelectedBadgeIndex(undefined);
   };
   const addItem = async () => {
-    if ((!!currentFieldValue || currentFieldValue === false) && !hasError) {
+    if (
+      (currentFieldValue !== undefined ||
+        currentFieldValue !== null ||
+        currentFieldValue !== \\"\\") &&
```
SwaySway, line 736: we can use double not to check for falsy values

letsbelopez: I want to allow a false here. It was a double not previously.

letsbelopez: I want to allow a number 0 as well.

SwaySway (approved):

hein-j (approved):

PR #687, opened 2022-10-07
fix: fix value props on ArrayField child

hein-j:
*Description of changes:*
_Problems:_
- `SwitchField` child of `ArrayField` had 2 `value` props.
- child of `ArrayField` always had `value` prop, even if the component did not need to be controlled. It would cause issues like TextField tied to AWSTimestamp DataStore field resetting to undefined `onChange`.
- child of `ArrayField` always had `value` prop even when it needed to be something else to be effective, like `isPressed`, `isChecked`, etc.

_Changes:_
- Change `value` on `SwitchField` to `isChecked`. `value` has no effect on this component.
- Remove `value` on child of `ArrayField` on components that don't need it.
- For components that do need to be controlled, render the correct prop (e.g. `isPressed`), not always just `value`


Jshhhh (approved):
LGTM

cwoolum (approved):

PR #689, opened 2022-10-10
fix: arrayfield items prop fallback to empty array when record is null

Jshhhh:
*Issue #, if available:*

*Description of changes:*
- Fix ArrayField component when DS array type record value is null

### DIFF
--- a/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
+++ b/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
@@ -2353,7 +2353,7 @@ export default function NestedJson(props) {
     const cleanValues = { ...initialValues, ...initialData };
     setFirstName(cleanValues.firstName);
     setLastName(cleanValues[\\"last-Name\\"]);
-    setLastName1(cleanValues.lastName);
+    setLastName1(cleanValues.lastName ?? []);
     setCurrentLastName1Value(undefined);
     setBio(cleanValues.bio);
     setErrors({});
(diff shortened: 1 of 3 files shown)


packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
```diff
@@ -2036,7 +2036,7 @@ export default function NestedJson(props) {
         }}
         currentFieldValue={currentNicknamesValue}
         label={\\"nick-Names2\\"}
-        items={nicknames}
```
hein-j, line 2039: 1. shouldn't the variable name still be `nicknames`?
2. what do you think about doing this in the reset function instead? Something like `setNicknames(cleanedValue.nickNames ?? [])`. If too big of a change, maybe not worth it.

packages/codegen-ui-react/lib/utils/forms/array-field-component.ts
```diff
@@ -1062,7 +1065,14 @@ export const renderArrayFieldComponent = (
         ),
         factory.createJsxAttribute(
           factory.createIdentifier('items'),
-          factory.createJsxExpression(undefined, factory.createIdentifier(renderedFieldName)),
+          factory.createJsxExpression(
+            undefined,
+            factory.createBinaryExpression(
+              factory.createIdentifier('items'),
```
awinberg-aws, line 1071: should this be `renderedFieldName`?

awinberg-aws (changes requested):

packages/codegen-ui-react/lib/utils/forms/array-field-component.ts
Jshhhh: Yeah it should thanks

packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
Jshhhh: 1. Yes it should
2. Yeah thats is possible

hein-j (commented):
We don't have to have the fallback if `initialValues`, not `cleanValues` because that means it's not getting from DataStore. And we already set the initial value of ArrayFields to an empty array.

Jshhhh:
> We don't have to have the fallback if `initialValues`, not `cleanValues` because that means it's not getting from DataStore. And we already set the initial value of ArrayFields to an empty array.

The array type field returned from DataStore can be null so the array type field state would be set to `null`, which would break ArrayField component

hein-j (approved):

awinberg-aws (approved):

PR #696, opened 2022-10-12
fix: by default hide cancel

SwaySway:
*Issue #, if available:*
asana task: https://app.asana.com/0/1202185248783861/1203024227380062/f
*Description of changes:*
- hide cancel component generation by editing cta matrix
- hide onCancel variable and type via shouldIncludeCancel helper function

### DIFF
--- a/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
+++ b/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
@@ -1895,8 +1895,7 @@ function ArrayField({
   );
 }
 export default function NestedJson(props) {
-  const { onSubmit, onCancel, onValidate, onChange, overrides, ...rest } =
-    props;
+  const { onSubmit, onValidate, onChange, overrides, ...rest } = props;
   const { tokens } = useTheme();
   const initialValues = {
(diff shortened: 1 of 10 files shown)


cwoolum (approved):

packages/codegen-ui/lib/types/form/form-cta.ts
```diff
@@ -17,7 +17,7 @@ import { StudioFieldPosition } from './position';
 
 export type StudioFormCTAButton =
   | {
-      excluded: true;
+      excluded: boolean;
```
hein-j, line 20: can we keep this so that it's either excluded is `true` or excluded is not there at all? That's how it works for field configs in the `Common` package too. 

So in `map-cta.ts`, it would be 
`element === 'cancel' && !buttons?.cancel`
Meaning if it's `{}`, it does have an override and it should be rendered. That's how it works for field configs too. 
That way, in the UI, customers can include the cancel button without necessarily modifying the label, etc.

For your `shouldIncludeCancel` logic in codegen-ui-react, the same would apply.
Or maybe you can grab the info from the `buttonMatrix` instead if it's available

SwaySway: synced offline - updated pr with new changes

cwoolum (approved):

hein-j (approved):

PR #719, opened 2022-10-18
chore: wrap relationship fields in ArrayField

hein-j:
*Description of changes:*
- Wrap relationship fields in `ArrayField`
- Init snapshot testing for `HAS_ONE` relationship

Delimitations:
- Since `Autocomplete` doesn't exist yet, this will break E2E.
- Tasks to map props to the `Autocomplete` and `ArrayField` outstanding.


cwoolum (approved):
LGTM

packages/codegen-ui-react/lib/forms/form-renderer-helper.ts
```diff
@@ -55,6 +55,8 @@ import {
   setStateExpression,
 } from './form-state';
 
+export const shouldWrapInArrayField = (config: FieldConfigMetadata) => config.isArray || config.relationship;
```
SwaySway, line 58: nit: I wonder if we should start splitting up functionality into separate functions for better readability (ex. relationship helpers into their own files)

SwaySway (approved):
🚀

packages/codegen-ui-react/lib/forms/form-renderer-helper.ts
hein-j: I agree this form-renderer-helper file is getting unwieldy. Will take note to split this up in a separate PR.

PR #722, opened 2022-10-20
feat: add relationship suggestion query to forms

scottyoung:
*Issue #, if available:*

*Description of changes:*
- stores suggestion data in variable to be used in form
- imports models when needed for form suggestion data
- works with multiple relationships
- this just makes the initial query, next step would be to use this query result in the autocomplete component

### DIFF
--- a/packages/codegen-ui-react/lib/forms/react-form-renderer.ts
+++ b/packages/codegen-ui-react/lib/forms/react-form-renderer.ts
@@ -72,6 +72,7 @@ import {
   buildUpdateDatastoreQuery,
   runValidationTasksFunction,
   mapFromFieldConfigs,
+  buildRelationshipQuery,
 } from './form-renderer-helper';
 import {
   buildUseStateExpression,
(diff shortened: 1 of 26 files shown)


cwoolum (dismissed):
LGTM

packages/codegen-ui/example-schemas/forms/post-datastore-relationship-multiple.json
```diff
@@ -0,0 +1,70 @@
...
+  "name": "BookCreateForm",
+  "formActionType": "create",
+  "dataType": {
+    "dataSourceType": "DataStore",
+    "dataTypeName": "Book"
+  },
+  "fields": {
+    "primaryAuthor": {
+      "inputType": {
+        "type": "Autocomplete",
```
hein-j, line 11: nit: the two form schemas added are titled "post-" but use "Book" schemas

packages/codegen-ui-react/lib/forms/react-form-renderer.ts
```diff
@@ -414,6 +416,38 @@ export abstract class ReactFormTemplateRenderer extends StudioTemplateRenderer<
...
+
+    // datastore relationship query
+    /**
+      const authorRecords = useDataStoreBinding({
+        type: 'collection',
+        model: Author,
+      }).items;
+    */
+    Object.values(formMetadata?.fieldConfigs).forEach((config) => {
+      if (config.relationship) {
```
hein-j, line 428: A couple comments here:
1. In line 442 of this file, there is already a place where we are iterating fieldConfigs. It'd be great to do this here so we're not doing it once more. 
2. Breaking this out as a helper in a new file in form-renderer-helper may help with maintainability
3. In line 494 of this file, the models are already added to imports, so no need to do it again :)

hein-j: One more thing - instead of getActionIdentifier, it'd be good to use getRecordsName from form-state.ts so that the variable name stays in sync

packages/codegen-ui/example-schemas/forms/post-datastore-relationship-multiple.json
scottyoung: updated

packages/codegen-ui-react/lib/forms/react-form-renderer.ts
scottyoung: Thanks, updated all of these.  A couple of these things were introduced recently so I didn't see them when writing this.

SwaySway (approved):

packages/codegen-ui-react/lib/forms/react-form-renderer.ts
hein-j: Yea for sure. Thanks for updating

hein-j (approved):

PR #724, opened 2022-10-21
chore: modify ArrayField to support has-one relationships

hein-j:
*Description of changes:*
- Add props to the instantiation of ArrayField wrapping the Autogen component
- Handle new props in the ArrayField component

### DIFF
--- a/packages/codegen-ui-react/lib/forms/form-renderer-helper/form-state.ts
+++ b/packages/codegen-ui-react/lib/forms/form-renderer-helper/form-state.ts
@@ -31,6 +31,7 @@ import {
   PropertyName,
 } from 'typescript';
 import { lowerCaseFirst } from '../../helpers';
+import { isModelDataType, shouldWrapInArrayField } from './render-checkers';
 
 export const getCurrentValueName = (fieldName: string) => `current${capitalizeFirstLetter(fieldName)}Value`;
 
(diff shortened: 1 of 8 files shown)


cwoolum (approved):

packages/codegen-ui-react/lib/forms/form-renderer-helper/form-state.ts
```diff
@@ -240,47 +241,49 @@ export const resetStateFunction = (fieldConfigs: Record<string, FieldConfigMetad
...
+    if (!stateNames.has(stateName)) {
+      const accessExpression = isValidVariableName(stateName)
+        ? factory.createPropertyAccessExpression(
+            factory.createIdentifier(recordOrInitialValues),
+            factory.createIdentifier(stateName),
+          )
+        : factory.createElementAccessExpression(
+            factory.createIdentifier(recordOrInitialValues),
+            factory.createStringLiteral(stateName),
+          );
```
SwaySway, line 257: not for this pr but ive been noticing we've been replicating this logic in a lot of places 
i wonder we should unify it into one function which simply returns the variable name

SwaySway (approved):
🚀

packages/codegen-ui-react/lib/forms/form-renderer-helper/form-state.ts
hein-j: I just created a separate helper in this PR. I missed using it here. Will rev in the next PR

PR #726, opened 2022-10-24
chore: support not-model HAS_ONE relationships

hein-j:
*Description of changes:*
- support fields that are not of model type but has relationship

*Additional changes:*
Bug fixes like:
- DataStore query rendering twice when model used twice
- runValidationTasks need to use display value
- Autocomplete suggestions assume wrong shape of model records
- Helper for property access was wrong

### DIFF
--- a/packages/codegen-ui-react/lib/forms/form-renderer-helper/display-value.ts
+++ b/packages/codegen-ui-react/lib/forms/form-renderer-helper/display-value.ts
@@ -13,9 +13,22 @@
   See the License for the specific language governing permissions and
   limitations under the License.
  */
-import { FieldConfigMetadata, isValidVariableName } from '@aws-amplify/codegen-ui';
+import {
+  FieldConfigMetadata,
+  InvalidInputError,
+  isValidVariableName,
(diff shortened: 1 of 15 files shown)


cwoolum (approved):

packages/codegen-ui-react/lib/forms/form-renderer-helper/display-value.ts
```diff
@@ -13,96 +13,171 @@
...
+}): CallExpression {
+  const recordString = 'r';
+
+  const labelExpression = isModelType
+    ? factory.createBinaryExpression(
+        factory.createCallChain(
+          getElementAccessExpression('getDisplayValue', fieldName),
+          factory.createToken(SyntaxKind.QuestionDotToken),
+          undefined,
+          [factory.createIdentifier('r')],
```
Jshhhh, line 84: Nit: This 'r' should be `recordString` also?

hein-j: right, will bundle in next PR

PR #728, opened 2022-10-26
fix: alias model names collision with component types

Jshhhh:
*Issue #, if available:*

*Description of changes:*
- Alias model imports when colliding with imported components in same file

### DIFF
--- a/packages/codegen-ui-react/lib/helpers/index.ts
+++ b/packages/codegen-ui-react/lib/helpers/index.ts
@@ -13,4 +13,5 @@
   See the License for the specific language governing permissions and
   limitations under the License.
  */
+
 export const lowerCaseFirst = (input: string) => input.charAt(0).toLowerCase() + input.slice(1);
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -20,10 +20,10 @@ jobs:
         with:
           path: amplify-cli
           repository: aws-amplify/amplify-cli
-      - name: Setup Node.js LTS
+      - name: Setup Node.js LTS/gallium
         uses: actions/setup-node@v2
         with:
-          node-version: lts/*
+          node-version: lts/gallium
       - name: Build amplify-codegen-ui
         working-directory: amplify-codegen-ui
         run: |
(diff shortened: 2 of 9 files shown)


packages/codegen-ui-react/lib/helpers/index.ts
```diff
@@ -14,3 +32,16 @@
   limitations under the License.
  */
 export const lowerCaseFirst = (input: string) => input.charAt(0).toLowerCase() + input.slice(1);
+
+export const addModelAlias = (componentTypes: string[], importCollection: ImportCollection, model: string) => {
```
awinberg-aws, line 36: nit: the rest of our code shouldn't need to be aware of the fact that the model name is potentially aliased. This's bit of collision detection/resolution is being abstracted away and all the rest of the app needs to know is that this is the way to add a model to the import collection

packages/codegen-ui-react/lib/helpers/index.ts
```diff
@@ -14,3 +32,16 @@
...
+  if (componentTypes.includes(model) || isPrimitive(model)) {
+    let existingModelAlias = importCollection.getMappedAlias(ImportSource.LOCAL_MODELS, model);
+    if (!existingModelAlias) {
+      existingModelAlias = `${model}Aliased`;
```
awinberg-aws, line 40: prefer to suffix with `Model` as this is more in line with what a customer would write if found in this situation

packages/codegen-ui-react/lib/helpers/index.ts
Jshhhh: In other words, the `importCollection` itself should be able to handle collisions and that the helper function is not needed?

packages/codegen-ui-react/lib/helpers/index.ts
Jshhhh: I was leaning towards a suffix that is less likely to collide again

packages/codegen-ui-react/lib/helpers/index.ts
awinberg-aws: I was originally commenting about the naming of the helper, but you took it to the next logical conclusion and I agree that a helper shouldn't be needed.

packages/codegen-ui-react/lib/helpers/index.ts
awinberg-aws: How can you be certain that a model doesn't already collide when adding `Alias` to it?

Jshhhh: Yeah that is true as well. Ill add a short uuid like string to it

PR #731, opened 2022-10-28
feat: add handling for excluded sectional elements

rtpascual:
*Issue #, if available:*

*Description of changes:*
Updated type and helper methods to support excluded sectional elements similar to excluded fields.

### DIFF
--- a/packages/codegen-ui/lib/generate-form-definition/generate-form-definition.ts
+++ b/packages/codegen-ui/lib/generate-form-definition/generate-form-definition.ts
@@ -69,8 +69,8 @@ export function generateFormDefinition({
     .concat(
       Object.entries(form.sectionalElements).map(([elementName, elementConfig]) => ({
         name: elementName,
-        position: elementConfig.position,
-        excluded: false,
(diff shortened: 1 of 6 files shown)


packages/codegen-ui/lib/generate-form-definition/generate-form-definition.ts
```diff
@@ -69,8 +69,8 @@ export function generateFormDefinition({
     .concat(
       Object.entries(form.sectionalElements).map(([elementName, elementConfig]) => ({
         name: elementName,
-        position: elementConfig.position,
-        excluded: false,
+        position: 'position' in elementConfig ? elementConfig.position : undefined,
+        excluded: 'excluded' in elementConfig ? elementConfig.excluded : false,
```
cwoolum, line 73: I'm not sure if we should be stricter about the presence of the excluded flag. Ideally, it should only ever be present if the element actually is excluded. That could be confusing to customers though so let's just keep it as is.

cwoolum (dismissed):

PR #735, opened 2022-10-29
fix: allow empty string for url field

Jshhhh:
*Issue #, if available:*

*Description of changes:*
- if url field value is empty string fallback to undefined for datastore


cwoolum (approved):
Aaah, so DataStore allows undefined but not empty strings. Interesting!

awinberg-aws (approved):

PR #738, opened 2022-10-31
fix: forms checkbox to reflect correct model state

scottyoung:
*Issue #, if available:*

*Description of changes:*
Make CheckboxField a controlled component similar to ToggleButton so that the field properly renders a default state of "checked" when it's supposed to.

Update form setting checkbox to checked, and then reload form
https://user-images.githubusercontent.com/1263887/198681650-ed1fb596-4883-414c-a248-c793194a60fd.mov

Current experience, loading form with checkbox in "checked" state does not set it to checked properly
https://user-images.githubusercontent.com/1263887/198681863-de079ca6-ae7b-42fb-8124-3406ba92d164.mov

### DIFF
--- a/packages/codegen-ui/lib/__tests__/generate-form-definition/helpers/form-field.test.ts
+++ b/packages/codegen-ui/lib/__tests__/generate-form-definition/helpers/form-field.test.ts
@@ -469,13 +469,12 @@ describe('getFormDefinitionInputElement', () => {
     const config = {
       inputType: {
         type: 'CheckboxField',
-        defaultChecked: true,
       },
     };
 
     expect(getFormDefinitionInputElement(config)).toStrictEqual({
       componentType: 'CheckboxField',
-      props: { label: 'Label', name: 'fieldName', value: 'fieldName', defaultChecked: true },
+      props: { label: 'Label', name: 'fieldName', value: 'fieldName' },
     });
   });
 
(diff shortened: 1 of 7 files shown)


packages/codegen-ui/lib/__tests__/generate-form-definition/helpers/form-field.test.ts
```diff
@@ -469,13 +469,12 @@ describe('getFormDefinitionInputElement', () => {
...
       inputType: {
         type: 'CheckboxField',
-        defaultChecked: true,
       },
     };
 
     expect(getFormDefinitionInputElement(config)).toStrictEqual({
       componentType: 'CheckboxField',
-      props: { label: 'Label', name: 'fieldName', value: 'fieldName', defaultChecked: true },
+      props: { label: 'Label', name: 'fieldName', value: 'fieldName' },
```
cwoolum, line 477: There was a certain reason we were using `defaultChecked` but I don't remember why. Were you able to test both Create and Update forms to ensure there are no regressions?

scottyoung: Yeah I was very concerned with regression so I've tested both Create and Update forms with checkboxes.  In both scenarios, the initial checkbox state is shown correctly, and create/update actions contain the correct checkbox state.

cwoolum (approved):

Jshhhh (approved):

PR #752, opened 2022-11-09
chore: add golden file for update form with hasOne relationship

scottyoung:
*Issue #, if available:*

*Description of changes:*
Adds golden file for an Update form that uses a hasOne relationship


packages/codegen-ui-golden-files/lib/react/forms/datastore-update-form-with-has-one-relationship/BookUpdateForm.jsx
```diff
@@ -0,0 +1,391 @@
...
+  };
+
+  // fetch model data to fill form with preset values
+  const [bookRecord, setBookRecord] = React.useState(book);
+  React.useEffect(() => {
+    const queryData = async () => {
+      const record = id ? await DataStore.query(Book, id) : book;
+      setBookRecord(record);
+      // get existing relationship record
+      const existingAuthor = record ? await DataStore.query(Author, (a) => a.id.eq(record.authorId)) : primaryAuthor;
```
cwoolum, line 204: Should this fall back to `book.author`? Wouldn't `primaryAuthor` always be undefined here?

packages/codegen-ui-golden-files/lib/react/forms/datastore-update-form-with-has-one-relationship/BookUpdateForm.jsx
```diff
@@ -0,0 +1,391 @@
...
+        );
+        if (validationResponses.some((r) => r.hasError)) {
+          return;
+        }
+        if (onSubmit) {
+          modelFields = onSubmit(modelFields);
+        }
+        try {
+          const original = await DataStore.query(Book, id);
+          await DataStore.save(
```
cwoolum, line 269: This just saves without needing to do anything else with `primaryAuthor`?

cwoolum (commented):
Just a few questions

packages/codegen-ui-golden-files/lib/react/forms/datastore-update-form-with-has-one-relationship/schema.ts
```diff
@@ -0,0 +1,193 @@
...
+                },
+              },
+              displayValue: {
+                bindingProperties: {
+                  property: 'Author',
+                  field: 'name',
+                },
+              },
+            },
+          ],
```
zchenwei, line 44: Maybe a dumb question. Could you provide me with the context about how do we get this form definition. specifically this `valueMappings`?

packages/codegen-ui-golden-files/lib/react/forms/datastore-update-form-with-has-one-relationship/BookUpdateForm.jsx
```diff
@@ -0,0 +1,391 @@
...
+          onChange={(e) => {
+            // called when user is typing
+            let { value } = e.target;
+            if (errors.primaryAuthor?.hasError) {
+              runValidationTasks('primaryAuthor', value);
+            }
+            setCurrentPrimaryAuthorDisplayValue(value);
+            setCurrentPrimaryAuthorValue(undefined); // empty out prev selected option
+          }}
+          suggestions={authorRecords.map((r) => ({
```
zchenwei, line 348: We rename this to `options`, maybe want to update this prop as well?

packages/codegen-ui-golden-files/lib/react/forms/datastore-update-form-with-has-one-relationship/BookUpdateForm.jsx
```diff
@@ -0,0 +1,391 @@
...
+            id: r.id,
+            label: getDisplayValue['primaryAuthor']?.(record) ?? r.id,
+          }))}
+          onSuggestionSelect={({ id, label }) => {
```
zchenwei, line 352: And also this will simply be `onSelect` now

packages/codegen-ui-golden-files/lib/react/forms/datastore-update-form-with-has-one-relationship/BookUpdateForm.jsx
scottyoung: If it can't find an existing relationship record, it should fallback to undefined.  I'll update it  from `primaryAuthor` to  `undefined` to be clearer

packages/codegen-ui-golden-files/lib/react/forms/datastore-update-form-with-has-one-relationship/BookUpdateForm.jsx
scottyoung: I've updated the code slightly to `primaryAuthor: primaryAuthor?.id,` so that it captures the `id` in `modelFields` and updates correctly

packages/codegen-ui-golden-files/lib/react/forms/datastore-update-form-with-has-one-relationship/BookUpdateForm.jsx
scottyoung: Updated, thanks

packages/codegen-ui-golden-files/lib/react/forms/datastore-update-form-with-has-one-relationship/BookUpdateForm.jsx
scottyoung: Updated, thanks

packages/codegen-ui-golden-files/lib/react/forms/datastore-update-form-with-has-one-relationship/schema.ts
scottyoung: The `fields` part of _StudioForm_ definition defines what the form fields will be.  In this case, it will be an `AutoComplete` field with bound properties from the Author model.  In the case of a golden file, I'm not sure how useful it is to define a `StudioForm` object it.  The `Schema` definition listed below is more useful since it defines the relationship schema for this golden file.

cwoolum (approved):

packages/codegen-ui-golden-files/lib/react/forms/datastore-update-form-with-has-one-relationship/BookUpdateForm.jsx
```diff
@@ -0,0 +1,394 @@
...
+    <Grid
+      as="form"
+      rowGap="15px"
+      columnGap="15px"
+      padding="20px"
+      onSubmit={async (event) => {
+        event.preventDefault();
+        let modelFields = {
+          name,
+          primaryAuthor: primaryAuthor?.id, // pass in id field
```
zchenwei, line 247: Wouldn't this be `authorId` according to the schema?

scottyoung: Good catch, it should be `authorId: primaryAuthor?.id`

zchenwei (approved):
Left one more question but not a blocker I think

cwoolum:
Have you tested this golden file in a local project to make sure it compiles and functions as expected?

packages/codegen-ui-golden-files/lib/react/forms/datastore-update-form-with-has-one-relationship/BookUpdateForm.jsx
```diff
@@ -0,0 +1,394 @@
...
+  React.useEffect(() => {
+    const queryData = async () => {
+      const record = id ? await DataStore.query(Book, id) : book;
+      setBookRecord(record);
+      // get existing relationship record
+      const existingAuthor = record ? await DataStore.query(Author, (a) => a.id.eq(record.authorId)) : undefined;
+      // set existing record on autocomplete if it exists
+      setPrimaryAuthor(existingAuthor);
+      setCurrentPrimaryAuthorValue(existingAuthor && existingAuthor[0].id);
+      setCurrentPrimaryAuthorDisplayValue(existingAuthor && existingAuthor[0]?.name);
```
zchenwei, line 208: These 2 are unnecessary IMO since it is not in edit and when users click on the badge and turn into edit state, they will be set.

scottyoung: They can also turn to edit state via the _Add item_ button, see recording: https://drive.corp.amazon.com/documents/scoyou@/Screen%20Recording%202022-11-16%20at%2010.19.09%20AM.mov

scottyoung:
> Have you tested this golden file in a local project to make sure it compiles and functions as expected?

Yes I've used a 1:1 copy of the golden file in my local project to test all the code, including the new Autocomplete component.

packages/codegen-ui-golden-files/lib/react/forms/datastore-update-form-with-has-one-relationship/BookUpdateForm.jsx
zchenwei: Right, but whats the point of populating something when you `Add`?

scottyoung: Yeah that makes sense, I'll update the golden file to remove those 2 lines

PR #771, opened 2022-11-18
fix: alias form model names so we do not get naming collisions

lyonsbp:
*Issue #, if available:*

*Description of changes:*
This changes handles the possibility of naming collisions within form relationships. Currently, the Studio Data Modeler will name the connected field `{ModelName}` with a capital letter when creating a `belongsTo` relationship. Since this name would conflict with the model name/import, a collision happened. This is resolved by aliasing the model import to `import {ModelName} as {ModelName}0`


zchenwei (approved):
LGTM

scottyoung (approved):

PR #777, opened 2022-11-19
Add display value support for enum arrays

letsbelopez:
*Issue #, if available:*

git push --set-upstream origin displayValueEnumArrays

*Description of changes:*

Show the display value text that we render for enums inside the badges of the array field.


packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
```diff
@@ -3436,6 +3436,537 @@ export default function MyPostForm(props: MyPostFormProps): React.ReactElement;
...
+  const [currentPostsValue, setCurrentPostsValue] = React.useState(undefined);
+  const PostsRef = React.createRef();
+  const [currentStatusesValue, setCurrentStatusesValue] =
+    React.useState(undefined);
+  const statusesRef = React.createRef();
+  const postRecords = useDataStoreBinding({
+    type: \\"collection\\",
+    model: Post,
+  }).items;
+  const getDisplayValue = {
```
cwoolum, line 3645: What happens if there are more than one array of enums? Should we add the field name in here to make it unique?

letsbelopez: If there's another array of enums, it will be added as a key value pair to this object. The key is the field name and the value is the function to return the display value.

PR #780, opened 2022-11-21
Add support for update forms with belongsTo relationship

lyonsbp:
*Issue #, if available:*

*Description of changes:*
This change adds support for belongsTo relationships in update forms by grabbing the nested model from the target record to update.

### DIFF
--- a/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
+++ b/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
@@ -3,8 +3,6 @@
 exports[`amplify form renderer tests datastore form tests custom form tests should render a custom backed form 1`] = `
 "/* eslint-disable */
 import * as React from \\"react\\";
-import { fetchByPath, validateField } from \\"./utils\\";
-import { getOverrideProps } from \\"@aws-amplify/ui-react/internal\\";
 import {
   Button,
   Flex,
@@ -15,6 +13,8 @@ import {
   StepperField,
   TextField,
 } from \\"@aws-amplify/ui-react\\";
+import { getOverrideProps } from \\"@aws-amplify/ui-react/internal\\";
+import { fetchByPath, validateField } from \\"./utils\\";
 export default function CustomDataForm(props) {
   const { onSubmit, onCancel, onValidate, onChange, overrides, ...rest } =
     props;
@@ -343,8 +343,8 @@ export default function CustomDataForm(props) {
 
 exports[`amplify form renderer tests datastore form tests custom form tests should render a custom backed form 2`] = `
 "import * as React from \\"react\\";
-import { EscapeHatchProps } from \\"@aws-amplify/ui-react/internal\\";
 import { GridProps, RadioGroupFieldProps, SelectFieldProps, StepperFieldProps, TextFieldProps } from \\"@aws-amplify/ui-react\\";
+import { EscapeHatchProps } from \\"@aws-amplify/ui-react/internal\\";
 export declare type ValidationResponse = {
     hasError: boolean;
     errorMessage?: string;
@@ -391,8 +391,6 @@ export default function CustomDataForm(props: CustomDataFormProps): React.ReactE
 exports[`amplify form renderer tests datastore form tests custom form tests should render a custom backed form 3`] = `
 "/* eslint-disable */
 import * as React from \\"react\\";
-import { fetchByPath, validateField } from \\"./utils\\";
-import { getOverrideProps } from \\"@aws-amplify/ui-react/internal\\";
 import {
   Button,
   Flex,
@@ -404,6 +402,8 @@ import {
   TextAreaField,
   TextField,
 } from \\"@aws-amplify/ui-react\\";
+import { getOverrideProps } from \\"@aws-amplify/ui-react/internal\\";
+import { fetchByPath, validateField } from \\"./utils\\";
 export default function CustomDataForm(props) {
   const {
     initialData,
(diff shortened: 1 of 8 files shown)


packages/codegen-ui-react/lib/forms/react-form-renderer.ts
```diff
@@ -447,6 +448,17 @@ export abstract class ReactFormTemplateRenderer extends StudioTemplateRenderer<
...
+      // Build effects to grab nested models off target record for relationships
+      Object.entries(formMetadata.fieldConfigs).forEach(([key, value]) => {
+        if (value.relationship) {
+          statements.push(
+            addUseEffectWrapper(buildGetRelationshipModels(key, lowerCaseDataTypeNameRecord), [
+              lowerCaseDataTypeNameRecord,
+            ]),
+          );
+        }
+      });
```
zchenwei, line 461: This works but I think it is not good when there are multiple `hasOne` relationship field. It will end up creating multiple `useEffect` hooks. I think there is a simple way to do the same is modify the above `buildUpdateDataStoreQuery`, passing `fieldConfigs` to it.

This is what it generates now
```
  React.useEffect(() => {
    const queryData = async () => {
      const record = id ? await DataStore.query(Member, id) : member;
      setMemberRecord(record);
    };
    queryData();
  }, [id, member]);
```

After adding your update it should be:
```
```
  React.useEffect(() => {
    const queryData = async () => {
      const record = id ? await DataStore.query(Member, id) : member;
      const team = record ? await record.team : undefined;
      setTeam(team);
      setMemberRecord(record);
    };
    queryData();
  }, [id, member]);
```
```

packages/codegen-ui-react/lib/forms/form-renderer-helper/relationship.ts
```diff
@@ -168,3 +168,55 @@ export const buildManyToManyRelationshipCreateStatements = (
...
+              undefined,
+              factory.createToken(SyntaxKind.EqualsGreaterThanToken),
+              factory.createBlock(
+                [
+                  factory.createIfStatement(
+                    factory.createIdentifier(recordName),
+                    factory.createBlock(
+                      [
+                        factory.createExpressionStatement(
+                          factory.createCallExpression(factory.createIdentifier(`set${fieldName}`), undefined, [
```
hein-j, line 195: It would be best to use the sanitized name instead of `fieldName`. I think there may be a helper to create a setState expression. It might be worth checking out how we're rendering `set___` in other places

packages/codegen-ui-react/lib/forms/react-form-renderer.ts
```diff
@@ -447,6 +448,17 @@ export abstract class ReactFormTemplateRenderer extends StudioTemplateRenderer<
           ['id', lowerCaseDataTypeName],
         ),
       );
+
+      // Build effects to grab nested models off target record for relationships
+      Object.entries(formMetadata.fieldConfigs).forEach(([key, value]) => {
+        if (value.relationship) {
```
hein-j, line 454: There's a place in this file that already iterates through fieldConfigs if you want to collapse this down there

packages/codegen-ui-react/lib/forms/react-form-renderer.ts
lyonsbp: I had it in a separate effect becuase there could be multiple, and I wasnt sure if having all of it in one effect would look messy. I went back and forth when implementing this, but. can change it to use the other effect

packages/codegen-ui-react/lib/forms/form-renderer-helper/relationship.ts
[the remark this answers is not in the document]
lyonsbp: Thanks, i'll check that out

packages/codegen-ui-react/lib/forms/react-form-renderer.ts
lyonsbp: I'll move this into the other effect as suggested by @zchenwei , so ill have iterate and pass the field configs into the `buildUpdateDatastoreQuery` function

zchenwei: Oh yeah, I think a conditional is required to check the the relationship type. For `hasMany` and `manyToMany`, the data lazy loading part is a bit different, it needs to call `toArray()`.

lyonsbp: Should I check for `belongsTo` and `hasOne` in that check then?

packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
```diff
@@ -5920,56 +5920,47 @@ export default function TagUpdateForm(props) {
...
-  const [label, setLabel] = React.useState(initialValues.label);
-  const [Posts, setPosts] = React.useState(initialValues.Posts);
+  const [name, setName] = React.useState(initialValues.name);
+  const [Team, setTeam] = React.useState(initialValues.Team);
   const [errors, setErrors] = React.useState({});
   const resetStateValues = () => {
-    const cleanValues = tagRecord
-      ? { ...initialValues, ...tagRecord, Posts: linkedPosts }
+    const cleanValues = memberRecord
+      ? { ...initialValues, ...memberRecord }
```
zchenwei, line 5931: Will the reset be a bit of problem? `cleanValues.Team` will be a promise then if simply spread `memberRecord`?

```
    const cleanValues = memberRecord
      ? { ...initialValues, ...memberRecord, Team }
```

and will have one more dependency

```
React.useEffect(resetStateValues, [memberRecord, Team]);
```

packages/codegen-ui-react/lib/forms/react-form-renderer.ts
```diff
@@ -462,9 +463,17 @@ export abstract class ReactFormTemplateRenderer extends StudioTemplateRenderer<
           ),
         );
       } else {
+        const relatedModelStatements: Statement[] = [];
+        // Build effects to grab nested models off target record for relationships
+        Object.entries(formMetadata.fieldConfigs).forEach(([key, value]) => {
+          if (value.relationship?.type === 'BELONGS_TO' || value.relationship?.type === 'HAS_ONE') {
+            // Flatten statments into 1d array
+            relatedModelStatements.push(...buildGetRelationshipModels(key));
```
hein-j, line 471: you are still passing in the field name instead of sanitized name || field name, so this will break if the field name happens to have hyphens, etc.
Also, the other feedback was that it could be great if you put this in the loop already in this file so we are not looping more times than necessary (line 490 I think?)

lyonsbp: Ah, the field name thing got lost in the rebase I did, ill re-add that back in. For the other concern, we can't really move it down to 490 because this needs to be inside the useEffect declared here.

zchenwei (approved):
LGTM

hein-j (approved):

PR #783, opened 2022-11-22
fix: nested boolean field isn't controlled

letsbelopez:
*Issue #, if available:*

*Description of changes:*

Give boolean fields the fully qualified name for the default value. 

For example, in the case of nested fields, it was setting isChecked to the nested object so it was passing an object `{}` to isChecked which was causing errors with the UI Component. UI Component expects a boolean or an undefined value.

### DIFF
--- a/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
+++ b/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
@@ -2157,6 +2157,7 @@ import {
   Heading,
   Icon,
   ScrollView,
+  SwitchField,
   Text,
   TextField,
   useTheme,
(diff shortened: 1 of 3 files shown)


packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
```diff
@@ -2671,6 +2690,35 @@ export default function NestedJson(props) {
         hasError={errors[\\"first Name\\"]?.hasError}
         {...getOverrideProps(overrides, \\"first Name\\")}
       ></TextField>
+      <SwitchField
+        label=\\"Enabled\\"
+        defaultChecked={false}
+        isChecked={options.enabled}
```
letsbelopez, line 2696: There was no existing test for boolean fields. Before the fix this would have been `isChecked={options}`

lyonsbp (approved):

cwoolum (approved):

PR #811, opened 2022-12-01
fix: return aliases with required models api

letsbelopez:
*Issue #, if available:*

https://app.asana.com/0/1202293561809020/1203468156027560/f

*Description of changes:*

When rendering component only, the client needs to know what the import alias names are for the required data models. This change adds the import aliases to the requiredDataModels value returned by the renderComponentOnly method, so the client knows which alias to use for the data model.

### DIFF
--- a/packages/codegen-ui-react/lib/forms/react-form-renderer.ts
+++ b/packages/codegen-ui-react/lib/forms/react-form-renderer.ts
@@ -118,7 +118,7 @@ export abstract class ReactFormTemplateRenderer extends StudioTemplateRenderer<
 
   public fileName: string;
 
-  protected requiredDataModels: string[] = [];
+  protected requiredDataModels: Map<string, { importAlias: string | undefined }> = new Map();
 
   protected shouldRenderArrayField = false;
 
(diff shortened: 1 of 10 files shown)


packages/codegen-ui-react/lib/forms/react-form-renderer.ts
```diff
@@ -314,8 +314,8 @@ export abstract class ReactFormTemplateRenderer extends StudioTemplateRenderer<
 
     // add model import for datastore type
     if (dataSourceType === 'DataStore') {
-      this.requiredDataModels.push(dataTypeName);
```
letsbelopez, line 317: I removed adding to `this.requiredDataModels` here because requiredDataModels is only used in `renderComponentOnly` and this function is not being called in `renderComponentOnly`. @lyonsbp can you double check this removal.

lyonsbp: yeah that should be fine

lyonsbp (approved):

packages/codegen-ui-react/lib/forms/react-form-renderer.ts
```diff
@@ -118,7 +118,7 @@ export abstract class ReactFormTemplateRenderer extends StudioTemplateRenderer<
 
   public fileName: string;
 
-  protected requiredDataModels: string[] = [];
+  protected requiredDataModels: Map<string, { importAlias: string | undefined }> = new Map();
```
Jshhhh, line 121: This will require changes in Studio. I feel we should keep it string[]

packages/codegen-ui-react/lib/forms/react-form-renderer.ts
Jshhhh: Im working on a task that will take care of this

hein-j (approved):

PR #821, opened 2022-12-07
chore: filter out selected options on Autocomplete

zchenwei:
*Issue #, if available:*

Selected options should not be available in the dropdown list. This PR will address the issue

Kapture 2022-12-07 at 10 10 32

*Description of changes:*

- Filter out selected options on Autocomplete
- Fix a conditional bug

### DIFF
--- a/packages/codegen-ui-react/lib/utils/forms/array-field-component.ts
+++ b/packages/codegen-ui-react/lib/utils/forms/array-field-component.ts
@@ -267,29 +267,7 @@ export const generateArrayFieldComponent = () => {
                 [
                   factory.createIfStatement(
                     factory.createBinaryExpression(
-                      factory.createParenthesizedExpression(
-                        factory.createBinaryExpression(
-                          factory.createBinaryExpression(
-                            factory.createBinaryExpression(
-                              factory.createIdentifier('currentFieldValue'),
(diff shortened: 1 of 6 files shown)


packages/codegen-ui-react/lib/utils/forms/array-field-component.ts
```diff
@@ -267,29 +267,7 @@ export const generateArrayFieldComponent = () => {
                 [
                   factory.createIfStatement(
                     factory.createBinaryExpression(
-                      factory.createParenthesizedExpression(
-                        factory.createBinaryExpression(
```
zchenwei, line 271: This condition - `(currentFieldValue !== undefined || currentFieldValue !== '')` is a bug. It will not rule out any nullish value :)

hein-j: I believe this was is verbose like this because we didn't want it to apply to `0`, `false`, etc.

zchenwei: Then should use `&`.

Jshhhh (approved):

Milan-Shah (approved):

PR #828, opened 2022-12-07
fix: make TextField a controlled component

scottyoung:
*Description of changes:*
Moving TextField from uncontrolled to controlled allows resetStateValues to properly reset its field to its initial state.  This is called both during Clear onClick as well as during a success Submit if clearOnSubmit is set to true.  This also contributes to the overall task of moving all components to controlled.

Tested locally in app with Date and Time fields, as well as interacting with onSubmit and reset onClick calls.

### DIFF
--- a/packages/codegen-ui/example-schemas/forms/bio-nested-create.json
+++ b/packages/codegen-ui/example-schemas/forms/bio-nested-create.json
@@ -52,7 +52,7 @@
         "below": "Nicknames1"
       }
     },
-    "bio.favoriteQuote": {
+    "bio.favorite Quote": {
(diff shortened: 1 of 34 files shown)


packages/codegen-ui/example-schemas/forms/bio-nested-create.json
```diff
@@ -61,23 +61,23 @@
         "below": "bio"
       }
     },
-    "bio.favorite-Animal": {
+    "bio.favoriteAnimal": {
```
Jshhhh, line 64: Why were these fields changed?

PR #838, opened 2022-12-14
feat: add getAliasMap method to importCollection

hein-j:
*Issue #, if available:*

*Description of changes:*
CMSv2 team is asking for an interface that would allow them to use just `importCollection` and get rid of dependency on `requiredDataModels`. This adds a method to `importCollection` that would let them do this easily.

### DIFF
--- a/packages/codegen-ui-react/lib/imports/import-collection.ts
+++ b/packages/codegen-ui-react/lib/imports/import-collection.ts
@@ -109,6 +109,19 @@ export class ImportCollection {
     ];
   }
 
+  getAliasMap(): { model: { [modelName: string]: string } } {
+    const modelMap: { [modelName: string]: string } = {};
+    const modelSet = this.#collection.get(ImportSource.LOCAL_MODELS);
+    const modelAliases = this.importAlias.get(ImportSource.LOCAL_MODELS);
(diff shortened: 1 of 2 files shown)


lyonsbp (approved):

packages/codegen-ui-react/lib/imports/import-collection.ts
```diff
@@ -109,6 +109,19 @@ export class ImportCollection {
...
+    const modelMap: { [modelName: string]: string } = {};
+    const modelSet = this.#collection.get(ImportSource.LOCAL_MODELS);
+    const modelAliases = this.importAlias.get(ImportSource.LOCAL_MODELS);
+    if (modelSet) {
+      [...modelSet].forEach((item) => {
+        const alias = modelAliases?.get(item);
+        if (alias) modelMap[item] = alias;
+      });
+    }
+    return { model: modelMap };
```
cwoolum, line 122: Why not just return `modelMap`?

hein-j: I wanted to make it extensible in case we wanted to return other aliases, such as that of custom Figma components.

cwoolum (approved):

PR #854, opened 2022-12-20
fix: load hasOne and hasMany relationships for collections the ds v4 way

hein-j:
*Description of changes:*
For collections, 
- load hasOne related records
- change the way hasMany related records are loaded - don't load all then filter, but instead, only load related records.
Tested e2e. Integ tests will be added in the `tagged-release/cms-support` to take advantage of models already in place there.

*Additional changes:*
- moved a few helpers out of form renderer to be shared with collections 


packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react.test.ts.snap
```diff
@@ -1147,13 +1182,20 @@ export default function CollectionWithModelNameCollisions(
   const buttonModelPagination = {
     sort: (s: SortPredicate<Button0>) => s.lastName(SortDirection.ASCENDING),
   };
+  const [buttonModel, setButtonModel] = React.useState(undefined);
   const buttonModelDataStore = useDataStoreBinding({
     type: \\"collection\\",
-    model: Button0,
+    model: Button,
```
Jshhhh, line 1188: This model should be aliased

hein-j: Thanks for the catch. Rev'd

scottyoung (approved):

Jshhhh (approved):

cwoolum (approved):

PR #861, opened 2022-12-28
fix: allow empty non-required fields to save

hein-j:
*Issue #, if available:*

*Description of changes:*
If a non-required field is empty, DataStore.save would fail because we would try to save `""` for `AWSEmail`, etc.

### DIFF
--- a/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
+++ b/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
@@ -379,6 +379,11 @@ export default function CreateCompositeDogForm(props) {
           modelFields = onSubmit(modelFields);
         }
         try {
+          Object.entries(modelFields).forEach(([key, value]) => {
+            if (typeof value === \\"string\\" && value.trim() === \\"\\") {
+              modelFields[key] = undefined;
+            }
+          });
           const compositeDog = await DataStore.save(
             new CompositeDog({
               ...modelFields,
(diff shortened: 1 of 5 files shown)


Jshhhh (dismissed):

packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
```diff
@@ -379,6 +379,11 @@ export default function CreateCompositeDogForm(props) {
           modelFields = onSubmit(modelFields);
         }
         try {
+          Object.entries(modelFields).forEach(([key, value]) => {
+            if (value === \\"\\") {
```
letsbelopez, line 383: Should we trim string in case the user presses the space bar? `if value.trim() === ""`

scottyoung (dismissed):

letsbelopez (changes requested):
Should we trim string in case the user presses the space bar? if value.trim() === ""

packages/test-generator/integration-test-templates/cypress/e2e/create-form-spec.cy.ts
```diff
@@ -92,7 +92,23 @@ describe('CreateForms', () => {
...
+          const record = JSON.parse(recordElement.text());
+
+          expect(record.string).to.equal('MyString');
+        });
+      });
+    });
+
     it('should save to DataStore', () => {
+      // reset state
+      cy.reload();
```
awinberg-aws, line 111: This implies to me that as it currently stands, these tests must execute in a specific order to pass. Is it possible to reset state between tests as part of a `beforeEach()`?

packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
hein-j: Makes sense. Will do

packages/test-generator/integration-test-templates/cypress/e2e/create-form-spec.cy.ts
hein-j: Good call, will do

letsbelopez (approved):

Jshhhh (approved):

PR #864, opened 2022-12-30
fix: fix updates in scalar fields

hein-j:
*Issue #, if available:*

*Issues addressed:*

- integer fields in update form - type some number (e.g 45); then backspace the entire number; click submit and the first digit will save (e.g. 4) instead of null

- integer, float, and json fields in update form - type some value (e.g 45); then highlight the entire number with your cursor and then remove the entire value with a single backspace. click submit and then the value is still saved (e.g. 45) instead of null

- float fields in update form - type some number (e.g. 78); then backspace the entire number, and click submit. the record is updated to include ‘0’ in the float field instead of null

- datetime and timestamp fields - if i backspace multiple times to delete an existing value in one of these fields and then click ‘submit,’ the original value is still saved instead of null.

### DIFF
--- a/packages/codegen-ui-react/lib/forms/form-renderer-helper/value-props.ts
+++ b/packages/codegen-ui-react/lib/forms/form-renderer-helper/value-props.ts
@@ -74,8 +74,12 @@ export const renderDefaultValueAttribute = (
     expression = factory.createJsxExpression(undefined, convertedValueAttributeMap[dataType](identifier));
   }
 
+  const componentsWithValue: string[] = ['TextField', 'TextAreaField'];
+
   return factory.createJsxAttribute(
-    componentType === 'TextField' ? factory.createIdentifier('value') : factory.createIdentifier('defaultValue'),
+    componentsWithValue.includes(componentType)
+      ? factory.createIdentifier('value')
+      : factory.createIdentifier('defaultValue'),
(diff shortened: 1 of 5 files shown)


scottyoung (approved):

packages/codegen-ui-react/lib/forms/form-renderer-helper/value-props.ts
```diff
@@ -74,8 +74,12 @@ export const renderDefaultValueAttribute = (
     expression = factory.createJsxExpression(undefined, convertedValueAttributeMap[dataType](identifier));
   }
 
+  const componentsWithValue: string[] = ['TextField', 'TextAreaField'];
```
letsbelopez, line 77: For things like this, can we use `Set` instead of `Array`. Set.has() is O(1) vs O(n) of an array. This is a really small list but we can start getting into the habit especially if codegen runs as a service at some point, performance is going to matter more.

letsbelopez (approved):

packages/codegen-ui-react/lib/forms/form-renderer-helper/value-props.ts
hein-j: Makes sense. Will do in a follow-up

PR #866, opened 2022-12-31
chore: change default label for related models

hein-j:
*Issue #, if available:*

*Description of changes:*
Details:
The format of the label should be "[firstNonIDField] - [primaryKey]"
If the first non-ID field is a string, it should be displayed as-is
If the customer's model has a composite primary key, the keys should be concatenated and separated by a hyphen

*Additional change (as separate commit):*
Change array of component names to operator per @letsbelopez 's comment


packages/codegen-ui-react/lib/forms/form-renderer-helper/value-props.ts
```diff
@@ -74,10 +74,10 @@ export const renderDefaultValueAttribute = (
     expression = factory.createJsxExpression(undefined, convertedValueAttributeMap[dataType](identifier));
   }
 
-  const componentsWithValue: string[] = ['TextField', 'TextAreaField'];
+  const componentsWithValue = new Set(['TextField', 'TextAreaField']);
```
letsbelopez, line 77: Since this has only two values, a regular old `if` statement is probably cool too.

letsbelopez (approved):

packages/codegen-ui-react/lib/forms/form-renderer-helper/value-props.ts
hein-j: Makes sense. pushing another rev

letsbelopez (approved):

packages/codegen-ui-react/lib/forms/form-renderer-helper/model-values.ts
```diff
@@ -316,36 +315,70 @@ export function getIDValueObject(idValueFunctions: PropertyAssignment[]) {
     ),
   );
 }
+// this is a measure to not show `undefined` the way `bquildConcatExpression` would if the first field is undefined
```
scottyoung, line 318: minor spelling nit

scottyoung (approved):

packages/codegen-ui-react/lib/forms/form-renderer-helper/model-values.ts
hein-j: thanks. taking note.

PR #868, opened 2023-01-03
fix: update imports and props for aliased components

rtpascual:
*Issue #, if available:*

*Description of changes:*
- Alias import to match aliased component type from original component. Example: `import { Grid as GridCustom } from "./Grid"`
- Point props import to original component - `import { GridProps } from "./Grid"` instead of `import { GridCustomProps } from "./GridCustom"`

### DIFF
--- a/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react.test.ts.snap
+++ b/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react.test.ts.snap
@@ -5454,17 +5454,19 @@ export default function CustomButton(
 exports[`amplify render tests custom components alias parent should render declarations 1`] = `
 "import * as React from \\"react\\";
 import { EscapeHatchProps } from \\"@aws-amplify/ui-react/internal\\";
-import { CustomButtonProps } from \\"./CustomButton\\";
+import { GridProps } from \\"./Grid\\";
+import { ButtonProps } from \\"@aws-amplify/ui-react\\";
 import { ViewTestProps } from \\"./ViewTest\\";
 export declare type PrimitiveOverrideProps<T> = Partial<T> & React.DOMAttributes<HTMLDivElement>;
-export declare type CustomParentAndChildrenOverridesProps = {
-    CustomParentAndChildren?: ViewTestProps;
-    MyCustomButton?: CustomButtonProps;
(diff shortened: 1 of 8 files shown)


hein-j (dismissed):

Jshhhh (dismissed):

packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react.test.ts.snap
```diff
@@ -5475,15 +5475,14 @@ import {
...
+import { Grid as GridCustom, GridProps } from \\"./Grid\\";
 import ViewTest, { ViewTestProps } from \\"./ViewTest\\";
-import { GridCustomProps } from \\"./GridCustom\\";
 
 export declare type PrimitiveOverrideProps<T> = Partial<T> &
   React.DOMAttributes<HTMLDivElement>;
 export declare type AliasParentOverridesProps = {
   AliasParent?: ViewTestProps;
-  Grid?: GridCustomProps;
+  Grid?: GridProps;
```
Jshhhh, line 5485: Is there a case where the primitive GridProps is used and a custom component GridProps?

rtpascual: I believe so, that would mean the user is using both the Grid primitive and a custom Grid component as children in the same parent, which seems unlikely in my opinion.

Would it be best to alias the custom component props as well like what was done for the component? i.e. `import { GridProps as GridCustomProps } from "./Grid"`

Jshhhh: Yeah, that would make sense. Can you add a test case for that?

rtpascual: Yeah sounds good, I'll have it in the next rev of this PR.

rtpascual: Updated validation test to be more specific for parent-child components with the same name and child components with the same name in new rev.

Jshhhh (approved):

hein-j (approved):

scottyoung (approved):

PR #870, opened 2023-01-04
fix: remove functions imported from devDep

hein-j:
*Issue #, if available:*

*Description of changes:*
Using functions imported from devDep `@aws-amplify/datastore` is causing CLI e2e tests to fail. 
This removes their usage and replaces w/ our own helper.

### DIFF
--- a/packages/codegen-ui-react/lib/forms/form-renderer-helper/render-checkers.ts
+++ b/packages/codegen-ui-react/lib/forms/form-renderer-helper/render-checkers.ts
@@ -14,7 +14,6 @@
   limitations under the License.
  */
 import { FieldConfigMetadata } from '@aws-amplify/codegen-ui';
-import { isEnumFieldType } from '@aws-amplify/datastore';
 
 export const shouldWrapInArrayField = (config: FieldConfigMetadata): boolean => config.isArray || !!config.relationship;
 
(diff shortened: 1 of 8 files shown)


Jshhhh (approved):

letsbelopez (approved):

packages/codegen-ui-react/lib/forms/form-renderer-helper/render-checkers.ts
```diff
@@ -23,8 +22,13 @@ export const isModelDataType = (
 ): config is FieldConfigMetadata & { dataType: { model: string } } =>
   !!(config.dataType && typeof config.dataType === 'object' && 'model' in config.dataType);
 
+export const isEnumDataType = (
+  config: FieldConfigMetadata,
+): config is FieldConfigMetadata & { dataType: { enum: string } } =>
+  !!(config.dataType && typeof config.dataType === 'object' && 'enum' in config.dataType);
```
awinberg-aws, line 28: prefer to check that `enum` is not undefined rather than just stopping at the property existing

packages/codegen-ui/lib/utils/form-component-metadata.ts
```diff
@@ -93,7 +92,11 @@ export const mapFormMetadata = (form: StudioForm, formDefinition: FormDefinition
           });
         }
 
-        if ((element.relationship || isEnumFieldType(element.dataType)) && 'valueMappings' in element) {
+        if (
+          (element.relationship ||
+            (element.dataType && typeof element.dataType === 'object' && 'enum' in element.dataType)) &&
```
awinberg-aws, line 97: can this reuse `isEnumDataType` in render-checker.ts?

hein-j: This code has been removed

packages/codegen-ui-react/lib/forms/form-renderer-helper/render-checkers.ts
hein-j: Finally addressing here - https://github.com/aws-amplify/amplify-codegen-ui/pull/991

PR #884, opened 2023-01-11
fix: add default value support for controlled textfield components

scottyoung:
*Description of changes:*
This is a follow up to moving TextField to a controlled component.  Since it's controlled, we can no longer use the defaultValue param, we have to pass in the default value to the initialValue object.

### DIFF
--- a/packages/codegen-ui-react/lib/helpers/index.ts
+++ b/packages/codegen-ui-react/lib/helpers/index.ts
@@ -13,7 +13,7 @@
   See the License for the specific language governing permissions and
   limitations under the License.
  */
-import { GenericDataField, GenericDataSchema } from '@aws-amplify/codegen-ui/lib/types';
+import { GenericDataField, GenericDataSchema, StudioFormFields } from '@aws-amplify/codegen-ui/lib/types';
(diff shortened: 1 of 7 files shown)


hein-j (approved):

packages/codegen-ui-react/lib/helpers/index.ts
```diff
@@ -113,3 +113,17 @@ export function isAliased(componentType: string): boolean {
 export function removeAlias(aliasString: string): string {
   return aliasString.replace(/(Custom$)/g, '');
 }
+
+export function getControlledComponentDefaultValue(
```
letsbelopez, line 117: I'm confused on this function. Can't a lot of components be controlled, but this one is only looking for TextField components?

scottyoung: There are other controlled components (like Toggle and Slider) but those seem to still support defaultValue as a component param.  Since I converted TextField to a controlled component, this is a direct follow up to that change which I confirmed does not support defaultValue.

Jshhhh (approved):

PR #886, opened 2023-01-18
fix: throw validation error when updating models with misconfigured h…

bombguy:
Updated behavior to throw before making DataStore calls when unlinking child model of hasMany relationship when the linking field is set as required.

### DIFF
--- a/packages/codegen-ui-golden-files/lib/react/forms/datastore-update-form-with-has-many-relationship/SchoolUpdateForm.jsx
+++ b/packages/codegen-ui-golden-files/lib/react/forms/datastore-update-form-with-has-many-relationship/SchoolUpdateForm.jsx
@@ -164,6 +164,7 @@ export default function SchoolUpdateForm(props) {
 
   const [schoolRecord, setSchoolRecord] = React.useState(school);
   const [linkedStudents, setLinkedStudents] = React.useState([]);
+  const canUnlinkStudents = false;
 
   React.useEffect(() => {
     const queryData = async () => {
@@ -252,6 +253,9 @@ export default function SchoolUpdateForm(props) {
           const studentsToUnLink = [];
           const studentsSet = new Set();
           const linkedStudentsSet = new Set();
+          if (!canUnlinkStudents && studentsToUnLink.length > 0) {
+            throw Error(`${original.id} cannot be unlinked from School because schoolID is a required field.`);
+          }
           Students.forEach((r) => studentsSet.add(r.id));
           linkedStudents.forEach((r) => linkedStudentsSet.add(r.id));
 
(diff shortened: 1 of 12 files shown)


packages/codegen-ui-golden-files/lib/react/forms/datastore-update-form-with-has-many-relationship/SchoolUpdateForm.jsx
```diff
@@ -252,6 +253,9 @@ export default function SchoolUpdateForm(props) {
           const studentsToUnLink = [];
           const studentsSet = new Set();
           const linkedStudentsSet = new Set();
+          if (!canUnlinkStudents && studentsToUnLink.length > 0) {
+            throw Error(`${original.id} cannot be unlinked from School because schoolID is a required field.`);
```
hein-j, line 257: Not quite sure what "original" is?
Shouldn't the ID coming from studentsToUnlink[0]?
Also, if it's easy enough, suggesting "Student" in the front - 
Student ${studentsToUnlink[0].id} cannot be unlinked from School because schoolID is a required field.`

packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
```diff
@@ -3765,27 +3767,19 @@ export default function UpdateCompositeDogForm(props) {
...
-              promises.push(
-                DataStore.save(
-                  CompositeToy.copyOf(original, (updated) => {
-                    updated.compositeDogCompositeToysName = null;
-                    updated.compositeDogCompositeToysDescription = null;
-                  })
-                )
+            if (!canUnlinkCompositeToys && compositeToysToUnLink.length > 0) {
+              throw Error(
+                \`CompositeToy \${original.id} cannot be unlinked from CompositeDog because compositeDogCompositeToysName is a required field.\`
```
hein-j, line 3772: Also, not every model has "id". You can look at the `primaryKeys` array attached to each model and just take the first one -- should be compositeToysToUnlink[0].kind or something.

bombguy: Got it. Will address

packages/codegen-ui-golden-files/lib/react/forms/datastore-update-form-with-has-many-relationship/SchoolUpdateForm.jsx
bombguy: ahh this one is on the golden file. I will move it over to the correct place. But the snapshot should have it in correct place

hein-j (approved):

Jshhhh (approved):

PR #899, opened 2023-01-23
fix: allow required arrays in relationships

hein-j:
*Issue #, if available:*

*Description of changes:*
When a customer has a `!` next to their array for `hasMany` or `manyToMany` relationships, forms will throw because we are explicitly setting them to `undefined`.
This PR excludes the fields instead, allowing the required array fields to not throw.
Note that the `!` next to the array does not actually have any meaning for the relationship itself, whatever the customer's intent is.

Q: Why do we have to exclude the fields? Why not just leave in `[]`?
A: For models with CPKs, leaving in `[]` will throw.

*Additional changes:*
Consolidate logic in the CTA DS call renderer so that it's more maintainable. 

### DIFF
--- a/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
+++ b/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
@@ -1675,12 +1675,14 @@ export default function CreateCompositeDogForm(props) {
               modelFields[key] = undefined;
             }
           });
+          const modelFieldsToSave = {
+            name: modelFields.name,
+            description: modelFields.description,
+            CompositeBowl: modelFields.CompositeBowl,
+            CompositeOwner: modelFields.CompositeOwner,
+          };
           const compositeDog = await DataStore.save(
-            new CompositeDog({
-              ...modelFields,
-              CompositeToys: undefined,
-              CompositeVets: undefined,
-            })
+            new CompositeDog(modelFieldsToSave)
           );
           const promises = [];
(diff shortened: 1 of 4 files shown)


packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
```diff
@@ -1675,12 +1681,13 @@ export default function CreateCompositeDogForm(props) {
               modelFields[key] = undefined;
             }
           });
+          const modelFieldsToSave = {
```
cwoolum, line 1684: Can we flip this to create a new object from the fields we want rather than deleting the fields we don't?

packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
```diff
@@ -896,7 +899,10 @@ export default function CreateCommentForm(props) {
               modelFields[key] = undefined;
             }
           });
-          await DataStore.save(new Comment(modelFields));
+          const modelFieldsToSave = {
```
cwoolum, line 902: Is this necessary?

packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
hein-j: Done

cwoolum (approved):

Jshhhh (approved):

PR #924, opened 2023-02-16
fix: add composite key type def

letsbelopez:
*Issue #, if available:*

*Description of changes:*

If there's a composite key, add the type def for it to the component props.

### DIFF
--- a/packages/codegen-ui-react/lib/forms/form-renderer-helper/event-handler-props.ts
+++ b/packages/codegen-ui-react/lib/forms/form-renderer-helper/event-handler-props.ts
@@ -54,21 +54,25 @@ import { getOnChangeValidationBlock } from './validation';
 import { buildModelFieldObject } from './model-fields';
 import { isModelDataType, shouldWrapInArrayField } from './render-checkers';
 import { extractModelAndKeys, getMatchEveryModelFieldCallExpression } from './model-values';
+import { COMPOSITE_PRIMARY_KEY_PROP_NAME } from '../../utils/constants';
 
-export const buildMutationBindings = (form: StudioForm, primaryKey?: string) => {
+export const buildMutationBindings = (form: StudioForm, primaryKeys: string[] = []) => {
   const {
     dataType: { dataSourceType, dataTypeName },
     formActionType,
   } = form;
   const elements: BindingElement[] = [];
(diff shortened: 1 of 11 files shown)


packages/codegen-ui-react/lib/forms/form-renderer-helper/event-handler-props.ts
```diff
@@ -54,21 +54,25 @@ import { getOnChangeValidationBlock } from './validation';
...
 
-export const buildMutationBindings = (form: StudioForm, primaryKey?: string) => {
+export const buildMutationBindings = (form: StudioForm, primaryKeys: string[] = []) => {
   const {
     dataType: { dataSourceType, dataTypeName },
     formActionType,
   } = form;
   const elements: BindingElement[] = [];
-  if (dataSourceType === 'DataStore' && primaryKey) {
+  if (dataSourceType === 'DataStore' && primaryKeys && primaryKeys.length) {
```
Jshhhh, line 65: Could just check for `.length` for primaryKeys since it defaults to []

letsbelopez: yes

packages/test-generator/integration-test-templates/src/FormTests/DSCompositeDog.tsx
```diff
@@ -157,12 +145,9 @@ export default function () {
       <Heading>DataStoreFormUpdateCompositeDog</Heading>
       <View id="DataStoreFormUpdateCompositeDog">
         <DataStoreFormUpdateCompositeDog
-          compositeDog={dataStoreFormUpdateCompositeDogRecord}
+          id={compositeDogPrimaryKey}
           onSuccess={async () => {
-            const record = (await DataStore.query(CompositeDog, {
-              name: 'Yundoo',
-              description: 'tiny but mighty',
```
hein-j, line 164: Instead of modifying the existing test, can you add a new one?
So, just render this component below with a different `id`, like `DataStoreFormUpdateCompositeDog-byID` and add the test in the spec making sure the existing data shows up.

letsbelopez: ok. I modified this one because I thought there were plenty of test cases for passing in a model

PR #939, opened 2023-03-08
fix: clear button on create form should reset select field

scottyoung:
## Problem
The Clear button on a Create Form does not reset a SelectField correctly.

## Solution
Setting the value of a SelectField to an empty string `""` rather than `undefined` will correctly set the value of the field to the first option of the dropdown, either the Placeholder, or the first option if no placeholder exists.

### Ticket

GitHub issue _____

### Manual tests
1. Load a Create form with a SelectField.  Choose an option for the SelectField, and then press the Clear button.  SelectField returns to placeholder text.
2. Load an Update form with a SelectField.  Choose a new option for the SelectField, and then press the Reset button.  SelectField returns to original option.

### Automated tests
- [x] E2E tests added/updated


packages/test-generator/integration-test-templates/cypress/e2e/form/DSAllSupportedFormFields-spec.cy.ts
```diff
@@ -443,22 +443,22 @@ describe('FormTests - DSAllSupportedFormFields', () => {
         const record = JSON.parse(recordElement.text());
 
         expect(record.stringArray).to.deep.equal([]);
-        expect('int' in record).to.equal(false);
```
Jshhhh, line 446: Curious, why did this test need to be changed?

Jshhhh (approved):

PR #948, opened 2023-03-17
fix: nested json array uncontrolled to controlled component fix

letsbelopez:
## Problem

Custom JSON forms nested text field arrays could not update. Because the TextField was going from uncontrolled to controlled and would not update beyond their initial value. 

Broken generated code
```
const initialValues = {
	basics: {},
	// This is the nested object with the Text
	favoriteThings: {},
	active: false,
	enabled: false,
};

const [favoriteThings, setFavoriteThings] = React.useState(
	initialValues.favoriteThings
);

<ArrayField ...>
	<TextField
		label="Animals"
		value={favoriteThings["animals"]} <--- this value is undefined
```

Fixed generated code
```
<TextField
	label="Animals"
	value={currentFavoriteThingsAnimalsValue} <--- uses the currentArrayFieldValue instead
```

Pressing the clear button would throw an error for a function that was not defined.

## Solution

Use the current array value of the field for the TextField instead of a reference to the nested object.

In the resetState function, stop adding a function in there that's not needed and doesn't exist.

Fixes aws-amplify/amplify-studio#817

Happy path from docs site: 

### Ticket

GitHub issue _____

### Problem

https://user-images.githubusercontent.com/16825996/226070803-a0f45c8a-be9a-4806-843d-0ef56b4e73b5.mov

### Fixed

https://user-images.githubusercontent.com/16825996/226070811-0b777b66-b8c5-4d3f-8058-541666f37ea6.mov

### Automated tests
- [x] E2E tests added/updated

### DIFF
--- a/packages/test-generator/integration-test-templates/cypress/e2e/form/CustomNestedJSON-spec.cy.ts
+++ b/packages/test-generator/integration-test-templates/cypress/e2e/form/CustomNestedJSON-spec.cy.ts
@@ -21,14 +21,19 @@ describe('FormTests - CustomNestedJSON', () => {
     cy.visit('http://localhost:3000/form-tests/CustomNestedJSON');
   });
 
-  specify('create form should have a working ArrayField', () => {
+  specify('create form should have a working nested TextField array', () => {
     cy.get('#CustomFormCreateNestedJson').within(() => {
       cy.contains('Add item').click();
       getDecoratedLabelSibling('Animals - optional').type('String1');
       cy.contains('Add').click();
       cy.contains('Add item').click();
       getDecoratedLabelSibling('Animals - optional').type('String2');
+      cy.contains('Add').click();
       cy.contains('String1').should('exist');
+      cy.contains('String2').should('exist');
+      cy.contains('Clear').click();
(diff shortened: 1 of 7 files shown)


packages/test-generator/integration-test-templates/cypress/e2e/form/CustomNestedJSON-spec.cy.ts
```diff
@@ -21,14 +21,19 @@ describe('FormTests - CustomNestedJSON', () => {
...
       getDecoratedLabelSibling('Animals - optional').type('String1');
       cy.contains('Add').click();
       cy.contains('Add item').click();
       getDecoratedLabelSibling('Animals - optional').type('String2');
+      cy.contains('Add').click();
       cy.contains('String1').should('exist');
+      cy.contains('String2').should('exist');
+      cy.contains('Clear').click();
+      cy.contains('String1').should('not.exist');
+      cy.contains('String2').should('not.exist');
```
letsbelopez, line 36: Tests that the TextField will actually update, and the clear button works.

Jshhhh (approved):

PR #973, opened 2023-03-30
chore: upgrade lerna to 6.4.1

rtpascual:
## Problem

We are 2 major versions behind for lerna.
## Solution

Bump up lerna from `^4.0.0` to `6.4.1` to be closer to the latest version.
### Ticket

GitHub issue _____

### Automated tests
- [ ] E2E tests added/updated
- [x] N/A - just a dev dependency bump


hein-j (approved):

package.json
```diff
@@ -51,7 +51,7 @@
     "eslint-plugin-prettier": "^3.4.0",
     "husky": ">=6",
     "jest": "^27.0.6",
-    "lerna": "^4.0.0",
+    "lerna": "6.4.1",
```
bombguy, line 54: any reason to keep it at specific version vs minimum version?

rtpascual: In lerna 6.5.0 the structure of the package.json was changed which breaks our commitlint config. There are open issues on lerna's Github that address this so we will wait for the fix then we can upgrade/use minimum version.

bombguy (approved):

PR #993, opened 2023-04-28
fix: update form with storage field should render with existing files

bombguy:
## Problem
Fixing update form with storage field where existing files are not rendered. StorageManager component does not update with `defaultFiles` prop update.

## Solution
Conditionally render after the record has been fetched.

### Automated tests
- [X] Snapshot updated to match updated golden file

### DIFF
--- a/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
+++ b/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
@@ -25391,6 +25391,7 @@ export default function UpdateProductForm(props) {
         errorMessage={errors.singleImgKey?.errorMessage}
         hasError={errors.singleImgKey?.hasError}
         label={\\"Single Image\\"}
+        descriptiveText={\\"Limited to One Image\\"}
         isRequired={false}
         isReadOnly={false}
       >
--- a/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
+++ b/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
@@ -25161,41 +25161,43 @@ export default function UpdateProductForm(props) {
         isRequired={false}
         isReadOnly={false}
       >
-        <StorageManager
-          defaultFiles={imgKeys.map((key) => ({ key }))}
-          onUploadSuccess={({ key }) => {
-            setImgKeys((prev) => {
-              let value = [...prev, key];
-              if (onChange) {
-                const modelFields = {
-                  name,
-                  imgKeys: value,
-                };
-                const result = onChange(modelFields);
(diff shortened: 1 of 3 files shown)


joebuono (approved):

packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
```diff
@@ -25161,41 +25161,43 @@ export default function UpdateProductForm(props) {
...
-              }
-              return value;
-            });
-          }}
-          accessLevel={\\"private\\"}
-          acceptedFileTypes={[\\".doc\\", \\".pdf\\"]}
-          isResumable={false}
-          showThumbnails={true}
-        ></StorageManager>
+        {productRecord && (
```
Jshhhh, line 25164: I feel like we should still render the storage field even if there is no productRecord. What is the reason why we want to hide the field if there is no record?

awinberg-aws: Does this have any impact on the usability of forms when no file has been uploaded yet? I'm not familiar with how we've implemented file uploads on forms but it looks like we expect StorageManager to do some form of event handling on upload and this update reads as if this element will not be on the page under certain conditions.

bombguy: @Jshhhh This was intentionally implemented this way. I had it rendered even when the record is undefined, but React seems to be comparing the DOM and linking the rendered component together. The StorageManager initializes the hook only when the component mount with defaultFiles prop.

bombguy: @awinberg-aws We are only blocking rendering StorageManager when the record is being queried. If you are updating a record without any files uploaded, it should not be blocking the render

Jshhhh (approved):

packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
bombguy: The StorageManager component uses `defaultFiles` to setup a reducer when the component mounts.
https://github.com/aws-amplify/amplify-ui/blob/main/packages/react-storage/src/components/StorageManager/hooks/useStorageManager/useStorageManager.tsx#L39

The hook is initialized only on mount and will not update on prop change. In order for the defaultFiles prop to work as intended, the prop needs to be defined on the first render of the component. This change is intended to block rendering until the record is fetched. 

I tried to conditionally render a different StorageManager when the record is not fetched yet. But due to React resolving the DOM between renders. It is treating the "unloaded" and "loaded" StorageManager to share reference. So the "loaded" StorageManager is being treated as just a prop change.

PR #995, opened 2023-05-01
feat: add processFile callback to rendered StorageManager for hashing filename

bombguy:
## Problem
When uploading files with same filename, they will overwrite the previously uploaded artifact.

## Solution
To reduce the number of possible naming conflict, generate a hash value from contents of the file. Uploading files with same content will generate same hash. It will save space in S3 bucket and reduce conflicts that will overwrite old files unintentionally. If the file contents are different, the hash value will be different as well.

### Automated tests
- [X] Snapshots updated to match intended change


packages/codegen-ui-react/lib/utils/constants.ts
```diff
@@ -16,3 +16,5 @@
 export const COMPOSITE_PRIMARY_KEY_PROP_NAME = 'id';
 
 export const STORAGE_FILE_KEY = 'key';
+
+export const STORAGE_FILE_ALGO_TYPE = 'SHA-1';
```
joebuono, line 20: Where is this being used?

joebuono (dismissed):

packages/codegen-ui-react/lib/utils/constants.ts
bombguy: Yup, updated in latest commit

packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
```diff
@@ -25476,6 +25506,21 @@ export default function UpdateProductForm(props) {
                 return value;
               });
             }}
+            processFile={async ({ file }) => {
+              const fileExtension = file.name.split(\\".\\").pop();
+              return file
```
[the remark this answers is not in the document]
bombguy: updated to be render in utils file

joebuono (approved):

rtpascual (approved):

PR #1006, opened 2023-05-03
feat: add graphql support for components

awinberg-aws:
## Problem

Customers who do not use DataStore are unable to build data-bound components

## Solution

Codegen has been updated with the ability to conditionally generate components using GraphQL APIs rather than using the DataStore library.

## Additional Notes

The update for model imports carries the assumption that the GraphQL API types (default path /src/API.ts) has been generated in the customer's project. These have no functional impact on the output and are just to help establish strict types in the customer project. If customers have the ability to codegen GraphQL operations without these types we will need followup work to gracefully fall back on using the `any` type.

### Ticket

GitHub issue _____

### Automated tests
- [ ] E2E tests added/updated

### DIFF
--- a/packages/codegen-ui-react/lib/__tests__/__utils__/amplify-renderer-generator.ts
+++ b/packages/codegen-ui-react/lib/__tests__/__utils__/amplify-renderer-generator.ts
@@ -156,3 +156,11 @@ export const renderExpanderJsxElement = (
 
   return transpile(expanderNode, {}).componentText;
 };
+
+export const rendererConfigWithGraphQL: ReactRenderConfig = {
+  apiConfiguration: {
+    dataApi: 'GraphQL',
+    operationsFilePath: '../graphql',
+    typesFilePath: '../API',
+  },
+};
--- a/packages/codegen-ui-react/lib/__tests__/imports/import-collection.test.ts
+++ b/packages/codegen-ui-react/lib/__tests__/imports/import-collection.test.ts
@@ -14,7 +14,7 @@
   limitations under the License.
  */
 import { ImportCollection, ImportSource } from '../../imports';
-import { assertASTMatchesSnapshot } from '../__utils__';
+import { assertASTMatchesSnapshot, rendererConfigWithGraphQL } from '../__utils__';
 
 function assertImportCollectionMatchesSnapshot(importCollection: ImportCollection) {
   assertASTMatchesSnapshot(importCollection.buildImportStatements());
@@ -70,7 +70,8 @@ describe('ImportCollection', () => {
       assertImportCollectionMatchesSnapshot(importCollection);
     });
     test('model imports colliding with exisiting imports', () => {
-      const importCollection = new ImportCollection({
+      const importCollection = new ImportCollection();
+      importCollection.ingestComponentMetadata({
         componentNameToTypeMap: { TextInput: 'TextField' },
         hasAuthBindings: false,
         requiredDataModels: [],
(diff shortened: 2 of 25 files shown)


packages/codegen-ui-react/lib/imports/import-collection.ts
```diff
@@ -35,11 +36,51 @@ export class ImportCollection {
...
+  addGraphqlQueryImport(importName: string) {
+    return this.addImport(
+      `${this.importConfig?.rendererConfig?.graphqlConfiguration?.docsFilePath}/queries`,
+      importName,
+    );
+  }
+
+  addGraphqlSubscriptionImport(importName: string) {
+    return this.addImport(
+      `${this.importConfig?.rendererConfig?.graphqlConfiguration?.docsFilePath}/subscriptions`,
```
cwoolum, line 69: Lot of duplication of the majority of this path. Can we clean it up?

packages/codegen-ui-react/lib/__tests__/studio-ui-codegen-react.test.ts
```diff
@@ -371,6 +381,26 @@ describe('amplify render tests', () => {
...
+        ).toMatchSnapshot();
+      });
+
+      it('DataStoreUpdateItem', () => {
+        expect(
+          generateWithAmplifyRenderer('workflow/dataStoreUpdateItem', rendererConfigWithGraphQL),
+        ).toMatchSnapshot();
+      });
+
+      it('DataStoreDeleteItem', () => {
```
cwoolum, line 397: Can we give better names to these tests? Are we just testing that we can codegen without an error?

packages/codegen-ui-react/lib/__tests__/imports/import-collection.test.ts
```diff
@@ -105,6 +106,88 @@ describe('ImportCollection', () => {
...
+      expect(importCollection.getMappedModelAlias('TextField0')).toEqual('TextField0');
+      expect(importCollection.getMappedModelAlias('ButtonProps')).toEqual('ButtonProps0');
+      expect(importCollection.getMappedModelAlias('React')).toEqual('React0');
+    });
+
+    // TODO: import collection currently does not alias import collisions
+    //   unless it's explicitly for the ImportSource.LOCAL_MODELS package
+    it('should successfully retrieve model alias when using GraphQL', () => {
+      const importCollection = new ImportCollection({ rendererConfig: rendererConfigWithGraphQL });
+      // importCollection.ingestComponentMetadata({
```
cwoolum, line 172: Can we clean up all the commented out pieces?

packages/codegen-ui-react/lib/imports/import-mapping.ts
```diff
@@ -22,6 +22,7 @@ export enum ImportSource {
   LOCAL_MODELS = '../models',
   LOCAL_SCHEMA = '../models/schema',
   UTILS = './utils',
+  AMPLIFY_API = '@aws-amplify/api',
```
cwoolum, line 25: Is the preference here to import from `@aws-amplify/api` or `aws-amplify`?

packages/codegen-ui-react/lib/workflow/action.ts
```diff
@@ -122,6 +157,111 @@ export function buildUseActionStatement(
...
+) {
+  // Import API client
+  const actionMutation = getGraphqlMutationForModel(action.action as Action, action.parameters.model);
+  importCollection.addMappedImport(ImportValue.API);
+  importCollection.addGraphqlMutationImport(actionMutation);
+  importCollection.addModelImport(action.parameters.model);
+
+  const inputKeys = [];
+
+  if ('id' in action.parameters && action.parameters.id) {
```
cwoolum, line 174: Should probably null check `action.parameters` as well unless we have validation upstream on the schema.

cwoolum (changes requested):

packages/codegen-ui-react/lib/workflow/action.ts
awinberg-aws: By this point the action has been narrowed to types that have a required parameters property.

packages/codegen-ui-react/lib/imports/import-mapping.ts
awinberg-aws: AppSync's documentation advises customers to install `@aws-amplify/api` and import from there. https://docs.aws.amazon.com/appsync/latest/devguide/building-a-client-app.html

I'll cut a task to add that package to our required packages check.

packages/codegen-ui-react/lib/__tests__/imports/import-collection.test.ts
awinberg-aws: I wanted to keep this as-is since the intention is to be able to fully uncomment this test once import collection has been updated to address the TODO above.

packages/codegen-ui-react/lib/__tests__/studio-ui-codegen-react.test.ts
awinberg-aws: These mirror the block of DataStore tests directly above them. The intent of these tests appear to be to verify that there is no error when rendering and to lock-in the snapshot as the expected output

cwoolum (dismissed):

packages/codegen-ui-react/lib/imports/import-collection.ts
awinberg-aws: this has been refactored to reduce duplication

rtpascual (dismissed):

rtpascual (dismissed):

packages/codegen-ui-react/lib/__tests__/__utils__/amplify-renderer-generator.ts
```diff
@@ -156,3 +156,11 @@ export const renderExpanderJsxElement = (
 
   return transpile(expanderNode, {}).componentText;
 };
+
+export const rendererConfigWithGraphQL: ReactRenderConfig = {
```
cwoolum, line 160: I think it might be better to take `dataApi` one level down so we contain everything related to that. 

```
{
  dataApi: 'GraphQL',
  operationsFilePath: '../graphql',
  typesFilePath: '../API',
}
```

or 

```
{
  dataApi: 'DataStore'
}
```

cwoolum (changes requested):

packages/test-generator/lib/generators/TestGenerator.ts
```diff
@@ -49,8 +50,8 @@ export type TestCase = {
 export type TestGeneratorParams = {
   writeToLogger: boolean;
   writeToDisk: boolean;
-  renderConfigOverride?: ReactRenderConfig;
-  outputConfigOverride?: ReactOutputConfig;
+  renderConfigOverride?: Partial<ReactRenderConfig>;
```
bombguy, line 53: Is this necessary to convert it to Partial? I'd prefer it to check for the required types when writing new tests. We already have DEFAULT_RENDER_CONFIG defined above. Can we not just add all the base config there for ease of use?

awinberg-aws: `DEFAULT_RENDER_CONFIG` did have the required property added to it, the issue is that the overrides should not be required to have all properties set which is why I had previously added the `Partial` to them. If you look at lines 66-67 where these override parameters are referenced you'll see that the use of `Partial` matches the intent as we spread the overrides over the defaults, but not having it was never a problem until recently because all the properties were optional.

packages/codegen-ui-react/lib/__tests__/__utils__/amplify-renderer-generator.ts
awinberg-aws: updated per suggestion

cwoolum (approved):

rtpascual (approved):

PR #1009, opened 2023-05-12
fix(codegen-ui-react): prevent useless props forward to custom components

JJK801:
## Problem
When we create composite components (components with nested components) in Figma and they go through the studio and amplify-codegen-ui, any props are forwarded to the nested component (including the defaults that were not configured in the studio).

When using variants with logic inside the components (breakpoint, displayMode, ...), this breaks the logic because the forwarded prop will always be the default variant.

Example:

GIVEN i created a NavBar component in Figma with 2 breakpoint variants (small, medium)
GIVEN i nested my NavBar component in a PageLayout component in Figma
WHEN codegen-amplify-ui generates my PageLayout component
THEN the nested NavBar component will receive a `breakpoint="small"` prop from PageLayout
THEN the forwarded prop overrides the breakpoint value generated inside the NavBar component

## Solution

Studio produced JSON includes a `configured` key for props added via the studio, so we can use it to filter custom components props. Implemented forwarding filter rule is the following:

- Component IS a Primitive => Because Primitives needs explicit props forwarding
- Component IS a HTML tag => Because HTML tags needs explicit props forwarding
- Prop value is dynamic (prop key is in `localStateReferences`)
- Prop value is a Collection binding
- Prop value is a concat
- Prop value has been configured in the studio

## Additional Notes
I fixed most of the specs by simply adding the `configured` prop to the dataset. It seems the best way to do, but pay attention so that doesn't change the nature of the test.

### Automated tests
- [ ] E2E tests added/updated

### DIFF
--- a/packages/codegen-ui-react/lib/react-component-renderer.ts
+++ b/packages/codegen-ui-react/lib/react-component-renderer.ts
@@ -20,6 +20,7 @@ import {
   StudioComponentChild,
   ComponentMetadata,
   StudioGenericEvent,
+  StudioComponentProperty,
 } from '@aws-amplify/codegen-ui';
 import {
   JsxAttributeLike,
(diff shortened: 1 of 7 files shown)


JJK801:
Hi 🖖 

Any news about this PR ?

PR #1013, opened 2023-05-19
fix: parse operand value when field is number type

Jshhhh:
## Problem
comparing equality with filter operand as string returns false. 0 == '0'

## Solution
Parse operand value when field type is a number

### Ticket

GitHub issue _____

### Automated tests
- [ ] E2E tests added/updated

### DIFF
--- a/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react.test.ts.snap
+++ b/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react.test.ts.snap
@@ -1094,7 +1094,7 @@ export default function CollectionOfCustomButtons(
   } = props;
   const itemsFilterObj = {
     and: [
-      { field: \\"age\\", operand: \\"10\\", operator: \\"gt\\" },
+      { field: \\"age\\", operand: 10, operator: \\"eq\\" },
       { field: \\"lastName\\", operand: \\"L\\", operator: \\"beginsWith\\" },
     ],
   };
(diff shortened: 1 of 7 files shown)


packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react.test.ts.snap
```diff
@@ -1094,7 +1094,7 @@ export default function CollectionOfCustomButtons(
   } = props;
   const itemsFilterObj = {
     and: [
-      { field: \\"age\\", operand: \\"10\\", operator: \\"gt\\" },
+      { field: \\"age\\", operand: 10, operator: \\"eq\\" },
```
cwoolum, line 1097: The test changed. Was this intentional?

cwoolum (approved):

packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react.test.ts.snap
Jshhhh: Yes. Expecting the operand to be number and not a string

packages/codegen-ui-react/lib/__tests__/react-component-render-helper.test.ts
```diff
@@ -345,4 +346,15 @@ describe('react-component-render-helper', () => {
       assertASTMatchesSnapshot(exp);
     });
   });
+
+  describe('parseNumberOperand', () => {
+    test('should parse int if field data type is Int', () => {
+      expect(parseNumberOperand('10', { dataType: 'Int', readOnly: false, required: false, isArray: false })).toBe(10);
+    });
+    test('should parse int if field data type is Int', () => {
```
rtpascual, line 354: nit: the test name should change to `Float`

Jshhhh: ah damn haha. ill fix it in another pr

PR #1020, opened 2023-06-06
fix: add getOverrideProps to StorageManager

bombguy:
## Problem
The override props are not being applied to StorageManager

## Solution
render spread attribute for getOverrideProps

### Ticket
GitHub issue : https://github.com/aws-amplify/amplify-studio/issues/970

### Automated tests
- [ ] Verified snapshot

### DIFF
--- a/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
+++ b/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
@@ -24984,6 +24984,7 @@ export default function CreateProductForm(props) {
         label={\\"Images\\"}
         isRequired={false}
         isReadOnly={false}
+        {...getOverrideProps(overrides, \\"imgKeys\\")}
       >
         <StorageManager
           onUploadSuccess={({ key }) => {
@@ -25212,6 +25213,7 @@ export default function UpdateProductForm(props) {
         label={\\"Images\\"}
         isRequired={false}
         isReadOnly={false}
+        {...getOverrideProps(overrides, \\"imgKeys\\")}
       >
         {productRecord && (
           <StorageManager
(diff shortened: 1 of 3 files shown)


Jshhhh (dismissed):

rtpascual (dismissed):

awinberg-aws (dismissed):

ErikCH (dismissed):

reesscot:
Shouldn't it also be possible to override the wrapping Field component?
```
      <Field
        errorMessage={errors.imgKeys?.errorMessage}
        hasError={errors.imgKeys?.hasError}
        label={\\"Images\\"}
        isRequired={false}
        isReadOnly={false}
      >
```

lyonsbp (dismissed):

Jshhhh:
> > Shouldn't it also be possible to override the wrapping Field component? I don't see an overrides call here:
> > ```
> >       <Field
> >         errorMessage={errors.imgKeys?.errorMessage}
> >         hasError={errors.imgKeys?.hasError}
> >         label={\\"Images\\"}
> >         isRequired={false}
> >         isReadOnly={false}
> >       >
> > ```
> 
> Yes, that is also possible as well. Should the overrideProps be spread in both Field & StorageManager?



> > Shouldn't it also be possible to override the wrapping Field component? I don't see an overrides call here:
> > ```
> >       <Field
> >         errorMessage={errors.imgKeys?.errorMessage}
> >         hasError={errors.imgKeys?.hasError}
> >         label={\\"Images\\"}
> >         isRequired={false}
> >         isReadOnly={false}
> >       >
> > ```
> 
> Yes, that is also possible as well. Should the overrideProps be spread in both Field & StorageManager?

I think the overrideProps should be one or the other, or have a different key passed into getOverrideProps

bombguy:
The override props should only be applied to the StorageManager since the configuration is specific to the StorageManager, not Field component.

This pattern is used in other wrapper component we render. (e.g. ArrayField)

Jshhhh (approved):

rtpascual (approved):

PR #1021, opened 2023-06-06
fix: cli test build

Jshhhh:
## Problem
cli test build fail on install

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -41,8 +41,7 @@ jobs:
       - name: Build amplify-cli
         working-directory: amplify-cli
         run: |
-          yarn policies set-version 1.18.0
-          yarn --network-concurrency 1
+          yarn dev-build
           yarn setup-dev
       - name: Install updated codegen libraries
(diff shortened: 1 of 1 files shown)


.github/workflows/check.yml
```diff
@@ -41,8 +41,7 @@ jobs:
       - name: Build amplify-cli
         working-directory: amplify-cli
         run: |
-          yarn policies set-version 1.18.0
-          yarn --network-concurrency 1
+          yarn dev-build
           yarn setup-dev
```
lazpavel, line 45: if this references the yarn setup-dev from CLI repo, it already does the build, yarn dev-build (was removed), you might want to add `yarn` before `yarn setup-dev`

.github/workflows/check.yml
```diff
@@ -52,8 +51,6 @@ jobs:
       - name: Build amplify-cli with updated codegen libraries
         working-directory: amplify-cli
         run: |
-          yarn policies set-version 1.18.0
-          yarn --network-concurrency 1
           yarn dev-build
```
lazpavel, line 54: this should be just `yarn build`

PR #1049, opened 2023-07-25
fix: change GraphQL unlink to use update action instead of delete

awinberg-aws:
## Problem

Unlinking child records in a hasMany relationship is currently using the delete operation for GraphQL forms. This is a destructive action and does not align with what is done on DataStore forms. Furthermore the form validation suggests the intended action is an update as it checks if the parent id field is required on the child.
## Solution

Update codegen to generate requests using the update operation to clear the parent reference on a child record.
### Ticket

GitHub issue:

### Automated tests

- [ ] E2E tests added/updated

### Housekeeping

- [ ] No non-essential console logs
- [ ] All new files contain license notice

### DIFF
--- a/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
+++ b/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
@@ -4232,7 +4232,7 @@ import { getOverrideProps } from \\"@aws-amplify/ui-react/internal\\";
 import { fetchByPath, validateField } from \\"./utils\\";
 import { API } from \\"aws-amplify\\";
 import { commentsByPostID, getPost, listComments } from \\"../graphql/queries\\";
-import { deleteComment, updateComment, updatePost } from \\"../graphql/mutations\\";
+import { updateComment, updatePost } from \\"../graphql/mutations\\";
 function ArrayField({
   items = [],
   onChange,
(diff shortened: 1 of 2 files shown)


packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
```diff
@@ -4620,10 +4620,11 @@ export default function PostUpdateForm(props) {
             }
             promises.push(
               API.graphql({
-                query: deleteComment,
+                query: updateComment,
```
cwoolum, line 4623: Shouldn't this be conditional based on if the relationship is required or not? Generating a form that could never succeed with a delete doesn't really make sense. The error message would be very vague as well, especially for the downstream customers.

awinberg-aws: We explicitly validate on if the relationship is required. if it is, then the update is prevented. This matches what is done today for forms using DataStore.

awinberg-aws: Discussed over slack and concluded that this achieves parity with DataStore forms but we should consider follow up work to improve the customer experience when working with child collections.

joebuono (approved):

bombguy (approved):

PR #1063, opened 2023-08-04
fix: remove trim on empty string replace, use null for replacement value

awinberg-aws:
## Problem
* properties are missing from create/update requests when non-required values are empty
* all-whitespace values are being trimmed on save but not when validated or passed into a customer's `onSubmit` callback.

## Solution
* update the empty string replacement operation to replace with `null` instead of `undefined` which was getting stripped when making the API request.
* remove string trimming when checking if a string is empty

### Ticket

GitHub issue:

### Automated tests

- [ ] E2E tests added/updated

### Housekeeping

### DIFF
--- a/packages/codegen-ui-react/lib/forms/form-renderer-helper/cta-props.ts
+++ b/packages/codegen-ui-react/lib/forms/form-renderer-helper/cta-props.ts
@@ -166,7 +166,7 @@ const getRecordUpdateDataStoreCallExpression = ({
 
 /**
   Object.entries(modelFields).forEach(([key, value]) => {
-    if (typeof value === 'string' && value.trim() === "") {
+    if (typeof value === 'string' && value === "") {
       modelFields[key] = undefined;
     }
   });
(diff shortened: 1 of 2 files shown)


rtpascual (approved):

packages/codegen-ui-react/lib/forms/form-renderer-helper/cta-props.ts
```diff
@@ -166,7 +166,7 @@ const getRecordUpdateDataStoreCallExpression = ({
 
 /**
   Object.entries(modelFields).forEach(([key, value]) => {
-    if (typeof value === 'string' && value.trim() === "") {
+    if (typeof value === 'string' && value === "") {
       modelFields[key] = undefined;
```
zchenwei, line 170: To my knowledge, need to update this comment as well?

zchenwei (approved):

packages/codegen-ui-react/lib/forms/form-renderer-helper/cta-props.ts
awinberg-aws: ah yeah, I'll update that comment on a future commit.

PR #1065, opened 2023-08-08
fix: use related fields for has_one and belongs_to properties when building GraphQL API payloads

awinberg-aws:
## Problem

When using GraphQL, fully hydrated properties are being sent on API requests leading to two problems:
* relationships are not being updated because the underlying related id is not being included on the request
* AppSync is rejecting the request because the hydrated property is not a part of the expected input type
## Solution

When building the model fields object, we iterate over all the configured fields. If the field has a relationship defined for it, we now use the target field of the relationship and the corresponding primary key on the related model type.
### Ticket

GitHub issue:

### Automated tests

- [ ] E2E tests added/updated

### Housekeeping

- [ ] No non-essential console logs
- [ ] All new files contain license notice


Jshhhh (approved):

packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
```diff
@@ -11058,9 +11058,9 @@ export default function CreateCommentForm(props) {
           if (onChange) {
             const modelFields = {
               name,
-              post,
-              User,
-              Org,
+              postID: post.id,
+              userCommentsId: User.id,
```
cwoolum, line 11062: Is this right? Why is that Pascal cased?

PR #1067, opened 2023-08-09
fix: update create parent and update child graphql expressions

rtpascual:
## Problem

Creating a record with child throws with the following error: `The variables input contains a field that is not defined for input object type '_____'`

## Solution

Get correct response data from create parent GraphQL statement and change inputs passed to update child GraphQL statement.

### Ticket

GitHub issue:

### Automated tests

- [ ] E2E tests added/updated

### Housekeeping

### DIFF
--- a/packages/codegen-ui-react/lib/forms/form-renderer-helper/relationship.ts
+++ b/packages/codegen-ui-react/lib/forms/form-renderer-helper/relationship.ts
@@ -265,6 +265,7 @@ export const buildManyToManyRelationshipStatements = (
   const joinTableThisModelName = relatedModelFields[0];
   const joinTableRelatedModelName = relatedJoinFieldName;
   const isGraphql = dataApi === 'GraphQL';
+  const isCompositeKey = thisModelPrimaryKeys.length > 1;
 
   if (!relatedJoinTableName) {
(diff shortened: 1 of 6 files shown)


packages/codegen-ui-react/lib/forms/form-renderer-helper/relationship.ts
```diff
@@ -1024,14 +1025,25 @@ export const buildManyToManyRelationshipStatements = (
...
-                              getCreateJoinTableExpression(
-                                relatedJoinTableName,
-                                savedModelName,
-                                joinTableThisModelName,
-                                joinTableRelatedModelName,
-                                importCollection,
-                                dataApi,
-                              ),
+                              isGraphql && isCompositeKey
+                                ? getGraphQLJoinTableCreateExpression(
```
awinberg-aws, line 1029: GraphQL version should be usable even when not composite key, reusing the composite key logic in there would allow this to be resilient to custom keys as well

rtpascual: Oh ok, going to look into supporting custom keys

awinberg-aws (approved):

Jshhhh (approved):

PR #1130, opened 2023-10-26
Fix undefined ref errors in {val}ToLinkMap callbacks

letsbelopez:
## Problem

There were errors when trying to reference a value in a callback function where the value was undefined because it was not referencing the correct variable name that was defined in the variable statements part of the components.

## Solution

I updated how the reference was created so it is referencing the correct var name.

## Additional Notes

In general I worked on simplifying the code of creating the `Records` value name to make sure it's consistent across each use case that needs that variable name.

### Ticket

GitHub issue:

### Manual tests

I updates all the snapshots and went through all the snapshots changes to make sure the references to all the name changes were pointing to the correct variable.

### Automated tests

- [ ] E2E tests added/updated

### Housekeeping

### DIFF
--- a/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
+++ b/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
@@ -194,7 +194,7 @@ export default function CreateOwnerForm(props) {
   const [name, setName] = React.useState(initialValues.name);
   const [Dog, setDog] = React.useState(initialValues.Dog);
   const [DogLoading, setDogLoading] = React.useState(false);
-  const [DogRecords, setDogRecords] = React.useState([]);
+  const [dogRecords, setDogRecords] = React.useState([]);
   const autocompleteLength = 10;
   const [errors, setErrors] = React.useState({});
   const resetStateValues = () => {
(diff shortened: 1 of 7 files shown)


packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
```diff
@@ -11579,7 +11579,7 @@ export default function MovieUpdateForm(props) {
             }
           });
           tagsToLinkMap.forEach((count, id) => {
-            const tagToLink = tagRecords.find((r) =>
+            const tagToLink = tagsRecords.find((r) =>
```
letsbelopez, line 11582: Here's a bug fix example. Notice how there's no corresponding diff for this variable name above this line. That's because tagRecords should have always been tagsRecords.

packages/codegen-ui-react/lib/forms/react-form-renderer.ts
```diff
@@ -586,12 +586,12 @@ export abstract class ReactFormTemplateRenderer extends StudioTemplateRenderer<
 
     // Add value state and ref array type fields in ArrayField wrapper
 
-    const relatedModelNames: Set<string> = new Set();
+    const relatedModelNames: Map<string, { relatedModelName: string; fieldName: string }> = new Map();
```
letsbelopez, line 589: Needed to also pass the field name

Milan-Shah (approved):

awinberg-aws (approved):

PR #1145, opened 2024-01-29
feat: pass form save result to onSuccess handler

Jshhhh:
## Problem

Fields object is passed into onSuccess handler in forms. The result of save call would be more helpful

## Solution

pass save result to onSuccess handler

### Manual tests

Manual save local

### Automated tests

- [x] E2E tests added/updated

### Housekeeping


packages/codegen-ui-react/lib/__tests__/react-forms/__snapshots__/form-renderer-helper.test.ts.snap
```diff
@@ -70,7 +70,7 @@ exports[`form-render utils should generate before & complete types if datastore
 "{
     clearOnSuccess?: boolean;
     onSubmit?: (fields: mySampleFormInputValues) => mySampleFormInputValues;
-    onSuccess?: (fields: mySampleFormInputValues) => void;
+    onSuccess?: (result: any) => void;
```
cwoolum, line 73: Not in love with the `any`. Don't we know the full shape of the object including Ids?

PR #1148, opened 2024-02-14
Dependabot upgrades

jacoblogan:
## Problem
Security vulnerabilities called out by dependabot

## Solution
Upgrade dependency versions

## Verification
Run both unit and integration tests

### Automated tests

- [ ] E2E tests added/updated
- [x] N/A - no new code added

### Housekeeping


packages/test-generator/package.json
```diff
@@ -29,12 +29,13 @@
     "@aws-amplify/datastore": "^4.0.0",
     "@aws-amplify/ui-react": "^5.0.0",
     "@aws-amplify/ui-react-storage": "^1.1.0",
-    "cypress": "12.0.2",
+    "cypress": "13.0.0",
```
jacoblogan, line 32: upgrading cypress to next major version 13 because we need to upgrade `@cypress/request` to >=3.0.0 which only works with cypress 13+.  This is being pinned to 13.0.0 because 13.0.0 supports node 16 even though it is being marked as deprecated but the current version 13.6.4 does not support node 16.  Reference https://docs.cypress.io/guides/references/changelog

zchenwei (approved):
LGTM

Jshhhh (approved):

PR #1156, opened 2024-02-21
fix: add optional chaining

jacoblogan:
## Problem
Codegened code is throwing this error `TypeError: undefined is not an object (evaluating 'record.ScanEvents.toArray')` when the field is undefined.

## Solution
Add optional chaining so that an error won't be thrown and instead the conditional empty array will be returned

## Verification
Tests are showing the desired optional chaining.

### Automated tests

- [ ] E2E tests added/updated

### Housekeeping

### DIFF
--- a/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
+++ b/packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
@@ -32139,7 +32139,7 @@ export default function UpdateOrgForm(props) {
     const queryData = async () => {
       const record = idProp ? await DataStore.query(Org, idProp) : orgModelProp;
       setOrgRecord(record);
-      const linkedComments = record ? await record.comments.toArray() : [];
+      const linkedComments = record ? await record?.comments?.toArray() : [];
       setLinkedComments(linkedComments);
     };
     queryData();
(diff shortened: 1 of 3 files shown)


packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
```diff
@@ -32139,7 +32139,7 @@ export default function UpdateOrgForm(props) {
     const queryData = async () => {
       const record = idProp ? await DataStore.query(Org, idProp) : orgModelProp;
       setOrgRecord(record);
-      const linkedComments = record ? await record.comments.toArray() : [];
+      const linkedComments = record ? await record?.comments?.toArray() : [];
```
Jshhhh, line 32142: This might return undefined which we don't want. Can you rewrite to 
```
const linkedComments = record && await record.comments?.toArray() || [];
```

packages/codegen-ui-react/lib/__tests__/studio-ui-codegen-react-forms.test.ts
```diff
@@ -228,7 +228,7 @@ describe('amplify form renderer tests', () => {
       expect(componentText).toContain('const studentRecords = useDataStoreBinding({');
 
       // check lazy load linked data
-      expect(componentText).toContain('const linkedStudents = record ? await record.Students.toArray() : [];');
+      expect(componentText).toContain('(record && (await record.Students?.toArray())) || [];');
```
jacoblogan, line 231: This line has grown long enough that it is split into 2 lines that is why we no longer see `const linkedStudents =` as part of the expect clause, but it is still in the generated code just on the line above this one.

packages/codegen-ui-react/lib/__tests__/__snapshots__/studio-ui-codegen-react-forms.test.ts.snap
jacoblogan: Thanks for the catch, I've gone through and updated this to coalesce to the empty array.

zchenwei (approved):

Jshhhh (approved):

PR #1170, opened 2025-04-09
fix(codegen-ui-react): add sanitization to binding expression

iCodeForBananas:


packages/codegen-ui-react/lib/react-component-render-helper.ts
```diff
@@ -128,21 +128,164 @@ export function isActionEvent(event: StudioComponentEvent): event is ActionStudi
...
+ *
+ * @param {string} str - The input string to validate
+ * @returns {string} Returns an empty string if dangerous patterns are found, otherwise returns the trimmed input string
+ *
+ * @example
+ * // Safe string
+ * validateString("Hello World"); // returns "Hello World"
+ *
+ * // Dangerous string
+ * validateString("eval('alert(1)')"); // returns ""
```
ioanabrooks, line 152: nit update fn name here

packages/codegen-ui-react/lib/react-component-render-helper.ts
```diff
@@ -128,21 +128,164 @@ export function isActionEvent(event: StudioComponentEvent): event is ActionStudi
...
+ */
+export function filterDangerousPatterns(str: string): string {
```
ioanabrooks, line 154: I like the renaming, my question regarding whether we're certain this will always be a string remains.

packages/codegen-ui-react/lib/react-component-render-helper.ts
iCodeForBananas: Gah I missed that, was just looking at the input and output.

packages/codegen-ui-react/lib/react-component-render-helper.ts
```diff
@@ -128,21 +128,164 @@ export function isActionEvent(event: StudioComponentEvent): event is ActionStudi
...
+  if (typeof str !== 'string') {
+    return '';
+  }
+
+  // Check for dangerous JavaScript patterns
+  const dangerousPatterns = [
```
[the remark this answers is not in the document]
iCodeForBananas: It could but currently this is the only place in the code base where it's relevant.

packages/codegen-ui-react/lib/react-component-render-helper.ts
iCodeForBananas: `const escapedPropertyValue = escapePropertyValue(property);`
This will always be a string because the value of property comes from the component schema, which is defined by the user.  That's where the XSS is injectable.

packages/codegen-ui-react/lib/react-component-render-helper.ts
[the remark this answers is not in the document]
ioanabrooks: But this way every time the fn is called we get a new object in memory

ioanabrooks:
Checks are failing due to `actions/upload-artifact@v3` being deprecated: https://github.com/actions/upload-artifact?tab=readme-ov-file#actionsupload-artifact

PR #1202, opened 2026-07-09
fix(codegen-ui-react): emit non-identifier theme keys as string literals

osama-rizk:
## Description

Theme keys are passed directly to `factory.createIdentifier(key)` in the theme renderer. `factory.createIdentifier()` emits its argument **verbatim** as source code, so a theme key that isn't a valid JavaScript identifier can produce malformed generated output.

This change routes theme keys through a small helper that:
- emits valid JS identifiers as identifiers (unchanged behavior), and
- emits any other key as a properly escaped **string-literal key**, keeping the generated object literal well-formed.

## Changes

- `react-theme-studio-template-renderer.ts`: add `buildThemePropertyName()` helper; use it at the four theme-key emission sites (`splitAndBuildThemeValues`, `buildThemeValues`, `buildThemeBreakpointValues`).
- Added a unit test covering a non-identifier theme key.

## Testing

- `react-theme-studio-template-renderer` suite: **7/7 pass**, all 6 existing snapshots unchanged (no behavior change for valid themes).
- `@aws-amplify/codegen-ui-react` builds clean (`tsc`).

### DIFF
--- a/packages/codegen-ui-react/lib/__tests__/react-theme-studio-template-renderer.test.ts
+++ b/packages/codegen-ui-react/lib/__tests__/react-theme-studio-template-renderer.test.ts
@@ -91,4 +91,23 @@ describe('react theme renderer tests', () => {
       expect(themeObject.breakpoints.defaultBreakpoint).toBe('base');
     });
   });
+
+  describe('theme keys that are not valid identifiers', () => {
+    it('should emit a non-identifier key as a quoted string-literal key', () => {
+      const nonIdentifierKey = 'foo("bar")';
+      const theme: StudioTheme = {
+        name: 'MyTheme',
+        values: [{ key: nonIdentifierKey, value: { value: 'red' } }],
+      };
+      const rendererFactory = new StudioTemplateRendererFactory(
+        (t: StudioTheme) => new ReactThemeStudioTemplateRenderer(t, {}),
+      );
+      const { componentText } = rendererFactory.buildRenderer(theme).renderComponent();
+
+      // The key is emitted as a quoted, escaped string-literal key, keeping the
+      // generated object literal well-formed rather than emitting it verbatim.
+      expect(componentText).toContain('\'foo("bar")\': "red"');
+      expect(componentText).not.toContain('foo("bar"): ');
+    });
+  });
 });
--- a/packages/codegen-ui-react/lib/__tests__/react-theme-studio-template-renderer.test.ts
+++ b/packages/codegen-ui-react/lib/__tests__/react-theme-studio-template-renderer.test.ts
@@ -13,11 +13,13 @@
   See the License for the specific language governing permissions and
   limitations under the License.
  */
+import { isIdentifier, isStringLiteral } from 'typescript';
 import { StudioTemplateRendererFactory, StudioTheme } from '@aws-amplify/codegen-ui';
 import { ScriptTarget, ScriptKind, ReactRenderConfig } from '..';
 import {
   ReactThemeStudioTemplateRenderer,
   ReactThemeStudioTemplateRendererOptions,
+  buildThemePropertyName,
 } from '../react-theme-studio-template-renderer';
 import { loadSchemaFromJSONFile } from './__utils__';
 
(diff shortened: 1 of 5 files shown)


packages/codegen-ui-react/lib/__tests__/react-theme-studio-template-renderer.test.ts
```diff
@@ -91,4 +91,23 @@ describe('react theme renderer tests', () => {
       expect(themeObject.breakpoints.defaultBreakpoint).toBe('base');
     });
   });
+
+  describe('theme keys that are not valid identifiers', () => {
+    it('should emit a non-identifier key as a quoted string-literal key', () => {
```
soberm: [blocker] The test payload `'foo("bar")'` is a malformed string, not a code-injection vector. A security patch that fixes an active CVE-class injection must prove that actual attack payloads are neutralized. Add explicit cases for the known bypass patterns, each asserting that the key appears only as a quoted property name and never as bare executable source:

- Object-literal breakout: `'}};process.exit(1);//'`
- Indirect eval: `'(0,eval)("x")'`
- String-concat obfuscation: `'"a"+"b"'`
- Null-byte smuggling: `"foo\x00bar"`

For example:
```typescript
it.each([
  ['object-literal breakout', '}};process.exit(1);//'],
  ['indirect eval',           '(0,eval)("x")'],
  ['string-concat',           '"a"+"b"'],
])('should not inject code for attack key: %s', (_, key) => {
  const theme: StudioTheme = { name: 'T', values: [{ key, value: { value: 'red' } }] };
  const { componentText } = rendererFactory.buildRenderer(theme).renderComponent();
  expect(componentText).toContain(`'${key}'`);
  expect(componentText).not.toMatch(new RegExp(`(?<!')${key.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')}(?!')`));
});
```

Additionally, `buildThemeValues` (nested token keys) and `buildThemeBreakpointValues` (breakpoint token keys) are not separately exercised with non-identifier inputs. Both delegate to `buildThemePropertyName`, but a future refactor that accidentally inlines the call differently would go undetected without a per-sink test.

packages/codegen-ui-react/lib/__tests__/react-theme-studio-template-renderer.test.ts
```diff
@@ -91,4 +91,23 @@ describe('react theme renderer tests', () => {
...
+      const nonIdentifierKey = 'foo("bar")';
```
soberm: [nitpick] `'foo("bar")'` mixes two non-identifier characters (parentheses and quotes) in a single payload, making it harder to see which character triggered the fallback path. Consider splitting into two cases: one with only a hyphen (`'font-size'`) to document the legitimate CSS-key path, and a clearly-labelled injection string for the security case.

packages/codegen-ui-react/lib/react-theme-studio-template-renderer.ts
```diff
@@ -47,6 +48,14 @@ export type ReactThemeStudioTemplateRendererOptions = {
   renderDefaultTheme?: boolean;
 };
 
+// Theme keys can be arbitrary strings, but factory.createIdentifier() emits its
```
soberm: [suggestion] The comment explains the mechanical behavior but omits the security rationale. A future maintainer evaluating whether to simplify back to `createIdentifier(key)` will find no in-code justification. Consider prepending:

```
// SECURITY: passing attacker-controlled StudioTheme keys directly to
// factory.createIdentifier() allows code injection in generated .tsx output
// (CVE-2025-4318 / H1 #3720454 -- theme-renderer surface).
```

packages/codegen-ui-react/lib/react-theme-studio-template-renderer.ts
```diff
@@ -47,6 +48,14 @@ export type ReactThemeStudioTemplateRendererOptions = {
...
+// argument verbatim as source. Emit only valid JS identifiers as identifiers;
+// any other key is emitted as a (properly escaped) string-literal key instead,
+// which keeps the generated object literal well-formed.
+function buildThemePropertyName(key: string): Identifier | StringLiteral {
```
soberm: [major] `buildThemePropertyName` is unexported, making it impossible to test directly. All coverage must flow through a full `renderComponent()` round-trip. For a security-critical predicate, export the function so a focused `describe('buildThemePropertyName')` block can enumerate the exact identity boundary -- `''`, `'$foo'`, `'--spacing-1'`, `'2xl'`, `'}};evil()'`, `'(0,eval)(...)'` -- without the cost and opacity of a full render cycle. If the regex is extracted to a shared constant (see the comment on the next line), the constant itself can be tested directly instead.

packages/codegen-ui-react/lib/react-theme-studio-template-renderer.ts
```diff
@@ -47,6 +48,14 @@ export type ReactThemeStudioTemplateRendererOptions = {
...
+  return /^[a-zA-Z_$][a-zA-Z0-9_$]*$/.test(key) ? factory.createIdentifier(key) : factory.createStringLiteral(key);
```
soberm: [major] This regex is now the third independent copy of a "valid JS identifier" pattern in this package. The same no-dot form also lives in `workflow/events.ts` (`buildBindingEvent`); a dot-allowing variant lives in `react-component-render-helper.ts` (`escapePropertyValue`). Three independent literals means a future hardening patch has three places to update and can silently diverge. Extract the no-dot form to a named constant in a shared utility and import it here and in `events.ts`:

```typescript
// packages/codegen-ui-react/lib/utils/identifiers.ts

/** Matches a simple JS identifier -- no dot-paths, no computed access.
 * SECURITY: values that fail this test must not be passed to factory.createIdentifier(). */
export const SIMPLE_JS_IDENTIFIER_RE = /^[a-zA-Z_$][a-zA-Z0-9_$]*$/;
```

packages/codegen-ui-react/lib/react-theme-studio-template-renderer.ts
```diff
@@ -47,6 +48,14 @@ export type ReactThemeStudioTemplateRendererOptions = {
...
+}
```
soberm, line 67: [suggestion] The fallback here emits the full original key as a quoted string literal. This is subtly different from the two existing sanitization patterns elsewhere in this package: `escapePropertyValue()` and `buildBindingEvent()` both fall back to `''`, producing `factory.createIdentifier('')` -- an empty identifier that causes a TypeScript compile error in the generated output. The string-literal approach is the sounder choice (valid syntax, and it supports legitimate non-identifier CSS keys like `'--spacing-1'` and `'2xl'`), but the divergence is silent. A brief inline comment documenting the intentional design would prevent a future reviewer from normalizing this to match the component-path pattern.

soberm (changes requested):

packages/codegen-ui-react/lib/__tests__/react-theme-studio-template-renderer.test.ts
osama-rizk: Done in 6d9614c. Replaced the single malformed-string case with explicit CVE-2025-4318-class payloads — object-literal breakout (`}};process.exit(1);//`), indirect eval (`(0,eval)("x")`) and string-concat obfuscation (`"a"+"b"`) — each asserting the key appears only as a quoted property name and never as a bare (unquoted) key via a negative-lookbehind regex. Also added dedicated per-sink tests exercising `buildThemeValues` (nested tokens) and `buildThemeBreakpointValues` (breakpoints) with a non-identifier input, so a future refactor that inlines the call can't silently regress.

packages/codegen-ui-react/lib/__tests__/react-theme-studio-template-renderer.test.ts
osama-rizk: Done. Split into a labelled legitimate CSS-key case (`font-size`, documenting the valid non-identifier path) and separate clearly-labelled injection cases.

packages/codegen-ui-react/lib/react-theme-studio-template-renderer.ts
osama-rizk: Done. Prepended the SECURITY rationale referencing CVE-2025-4318 (theme-renderer surface) so the justification is in-code.

packages/codegen-ui-react/lib/react-theme-studio-template-renderer.ts
osama-rizk: Done. `buildThemePropertyName` is now exported and covered by a focused `describe('buildThemePropertyName')` block enumerating the identity boundary (`''`, `$ref`, `2xl`, `--spacing-1`, `(0,eval)(...)`, object-literal breakout) directly, without a full render round-trip.

packages/codegen-ui-react/lib/react-theme-studio-template-renderer.ts
osama-rizk: Done. Extracted `SIMPLE_JS_IDENTIFIER_RE` to `lib/utils/identifiers.ts` with the SECURITY note, and imported it here and in `workflow/events.ts`. The no-dot literal is no longer duplicated across the two sinks.

packages/codegen-ui-react/lib/react-theme-studio-template-renderer.ts
osama-rizk: Done. Added an inline comment documenting that the string-literal fallback is intentional and differs from the `''` fallback in `escapePropertyValue()`/`buildBindingEvent()` — it keeps the object literal well-formed and preserves legitimate non-identifier CSS keys (`--spacing-1`, `2xl`) — with a 'do not normalize to '' ' note.

osama-rizk:
Thanks for the thorough review, @soberm — all six comments addressed in 6d9614c:

- **[blocker]** Replaced the malformed-string test with explicit CVE-class attack payloads (object-literal breakout, indirect eval, string-concat), each asserting quoted-key-only / never-bare emission. Added per-sink tests for `buildThemeValues` and `buildThemeBreakpointValues`.
- **[major] regex duplication** — extracted `SIMPLE_JS_IDENTIFIER_RE` to `lib/utils/identifiers.ts`, now imported by the theme renderer and `workflow/events.ts`.
- **[major] untestable helper** — exported `buildThemePropertyName`; added a direct `describe` block enumerating the identifier boundary.
- **[suggestion] security comment** — added the CVE-2025-4318 rationale in-code.
- **[suggestion] fallback divergence** — documented the intentional string-literal fallback vs the `''` fallback in the component path.
- **[nitpick]** — split the payload into a labelled legitimate CSS-key case and injection cases.

Verification: `react-theme-studio-template-renderer` 24/24 pass, `events` 21/21 pass, `tsc` build clean.

soberm (approved):

bobbor (approved):

